PR #16, opened 2022-09-13
Create Pods for an ElasticJob

workingloong:


operator/pkg/controllers/pod.go
```diff
@@ -0,0 +1,87 @@
...
+
+// PodManager manages the lifecycle of a pod including creation, updation and deletion.
+type PodManager struct{}
+
+func newPodManager() *PodManager {
+	return &PodManager{}
+}
+
+// CreatePod creates a Pod according to a PodTemplateSpec
+func (m *PodManager) CreatePod(job *elasticv1alpha1.ElasticJob, podTemplate *corev1.PodTemplateSpec, taskType TaskType, taskIndex int) *corev1.Pod {
```
samplise, line 45: I suggest changing "CreatePod" to "GeneratePod". In controller, "CreatePod" is more like creating a Pod in k8s cluster.

samplise (changes requested):

operator/pkg/controllers/pod.go
workingloong: Yeah!

samplise (approved):

PR #21, opened 2022-09-17
Update job status according to the status of master

workingloong:


operator/pkg/controllers/master.go
```diff
@@ -75,3 +83,61 @@ func (m *MasterManager) createPod(r *ElasticJobReconciler, job *elasticv1alpha1.
 	}
 	return nil
 }
+
+func (m *MasterManager) createPodName(job *elasticv1alpha1.ElasticJob) string {
```
samplise, line 87: I think `generatePodName` should be better.

operator/pkg/controllers/master.go
```diff
@@ -75,3 +83,61 @@ func (m *MasterManager) createPod(r *ElasticJobReconciler, job *elasticv1alpha1.
...
+		if job.Status.CompletionTime == nil {
+			now := metav1.Now()
+			job.Status.CompletionTime = &now
+		}
+		updateStatus(&job.Status, commonv1.JobFailed, reason, msg)
+	} else if master.Status.Phase == corev1.PodRunning || master.Status.Phase == corev1.PodPending {
+		job.Status.ReplicaStatuses[ReplicaTypeEasydlMaster].Active = 1
+		if !isRunning(job.Status) {
+			msg := fmt.Sprintf("job(%s/%s) is running.", job.Namespace, job.Name)
+			updateStatus(&job.Status, commonv1.JobRunning, commonutil.JobRunningReason, msg)
```
samplise, line 125: Does "!isRunning(job.Status)" indicate the job is pending? Why to update its status to running?

samplise (changes requested):

operator/pkg/controllers/master.go
workingloong: Done

operator/pkg/controllers/master.go
workingloong: I reuse `JobConditionType` from kubeflow and there is no `Pending`.So, I set `RUNNING` to job if the master pod is pending. We can add a `PENDING` status if necessary later.
https://github.com/kubeflow/common/blob/master/pkg/apis/common/v1/types.go#L112-L137

samplise (approved):

PR #26, opened 2022-09-23
Update job status with PS and worker Pods

### LINKED ISSUE #25: Create workers according to scaler

workingloong:
Fix #25

### DIFF
--- a/operator/pkg/controllers/psstrategy/worker.go
+++ b/operator/pkg/controllers/psstrategy/worker.go
@@ -38,7 +38,10 @@ func newWorkerManager() *WorkerManager {
 }
 
 func (m *WorkerManager) generateWorker(job *elasticv1alpha1.ElasticJob, replicaIndex int32) *corev1.Pod {
-	spec := job.Spec.ReplicaSpecs[ReplicaTypeWorker]
+	spec, ok := job.Spec.ReplicaSpecs[ReplicaTypeWorker]
+	if !ok {
+		return nil
+	}
 	name := m.generatePodName(job, replicaIndex)
 	pod := m.GeneratePod(job, &spec.Template, name)
 	pod.Labels[LabelRestartCount] = fmt.Sprintf("%d", spec.RestartCount)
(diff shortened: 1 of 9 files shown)


operator/pkg/controllers/psstrategy/worker.go
```diff
@@ -39,11 +38,12 @@ func newWorkerManager() *WorkerManager {
 }
 
 func (m *WorkerManager) generateWorker(job *elasticv1alpha1.ElasticJob, replicaIndex int32) *corev1.Pod {
-	workerSpec := job.Spec.ReplicaSpecs[ReplicaTypeWorker]
-	workerTemplate := &workerSpec.Template
-	workerName := m.generatePodName(job, replicaIndex)
-	pod := m.GeneratePod(job, workerTemplate, workerName)
-	pod.Labels[LabelRestartCount] = fmt.Sprintf("%d", workerSpec.RestartCount)
+	spec := job.Spec.ReplicaSpecs[ReplicaTypeWorker]
```
samplise, line 41: check if "ReplicaSpecs == nil" ?

operator/pkg/controllers/psstrategy/worker.go
```diff
@@ -39,11 +38,12 @@ func newWorkerManager() *WorkerManager {
...
+	name := m.generatePodName(job, replicaIndex)
+	pod := m.GeneratePod(job, &spec.Template, name)
+	pod.Labels[LabelRestartCount] = fmt.Sprintf("%d", spec.RestartCount)
```
samplise, line 44: Labels can be empty

operator/pkg/controllers/psstrategy/worker.go
workingloong: No, `ReplicaSpecs` has been defined as a map and it will be initialized. But, we need to check whether  ReplicaTypeWorker is in it.

operator/pkg/controllers/psstrategy/worker.go
workingloong: `GeneratePod` has check it.
https://github.com/intelligent-machine-learning/easydl/blob/1ea504b4a39e52279b02885ff5e8d77499f7b9d7/operator/pkg/controllers/pod.go#L44-L46

samplise (approved):

PR #43, opened 2022-10-08
Update README.md

linkerzhang:


linkerzhang:
The PR is trying to clarify the scope of EasyDL, which is critical before open sourcing it. EasyDL is not only doing resource configuration. It targets to take care of all engineering stuff in model training, starting from resource configuration (statically and dynamically), auto-distribution, etc.

workingloong (approved):
LGTM

PR #45, opened 2022-10-12
Add mysql db client

samplise:


workingloong:
We'd better not place vendors in the repo which will make the repo very huge.

brain/pkg/datastore/dbbase/db.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/DATA-DOG/go-sqlmock"
+	log "github.com/golang/glog"
+	"strings"
+	"time"
+	"xorm.io/xorm"
+	"xorm.io/xorm/names"
+)
+
+// DB is the struct of database
+type DB struct {
```
workingloong, line 28: It is better to use the full name DataBase.

brain/pkg/datastore/recorder/mysql/task_group_recorder.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/intelligent-machine-learning/easydl/brain/pkg/datastore/dbbase"
+	"time"
+	"xorm.io/xorm"
+)
+
+// tableTaskGroups is the name of job metrics table
+const tableTaskGroups = "task_groups"
+
+// TaskGroup represents a task structure
+type TaskGroup struct {
```
workingloong, line 26: Do we need TaskGroup or Task? I think the resources of each pod can be stored in the column of `resource` of job_metrics.

workingloong (approved):

PR #153, opened 2023-01-04
Add tensorflow failover/ps migration/checkpoint save hook/ model exporter

hxdtest:


hxdtest:
Add tensorflow failover/ps migration and it will be tested when kubernetes cluster is ready.
Add tensorflow checkpoint save hook and exporter.

dlrover/trainer/tensorflow/failover/tensorflow_failover.py
```diff
@@ -0,0 +1,165 @@
...
+                    task = i.tasks[0]
+                    if isinstance(task, dict):
+                        ind, address = list(task.items())[0]
+                    else:
+                        ind = 0
+                        address = task
+            logger.info(
+                "worker ind is {} and address is {}".format(ind, address)
+            )
+            for i in range(2):
```
workingloong, line 149: 2 is a magic number.
Can we replace 2 with`len(config.cluster_def.job)`?

hxdtest: 2 is a magic number. 
Fixed
```
      len_job = len(config.cluster_def.job)
      for i in range(len_job):
          # pop ps and worker/chief info
          config.cluster_def.job.pop()
```

And how we make sure that job.pop() pop "ps" and "worker" not "chief" or "evaluator"?
First of all there is a typo pop ps and worker info -> pop ps and worker/chief info

config.cluster_def is built in dlrover/trainer/estimator/base_estimator，the job contains ps info and either chief or worker info.  TensorflowFailover doesn't work for standalone evaluator.

workingloong (approved):
LGTM

PR #222, opened 2023-02-11
[Refactor] Implement an auto-scaler for PS Training

### LINKED ISSUE #217: Add to hook to check pre-requisities when executing scale plan.
Add to hook to check pre-requisities when executing scale plan. Because in different occasion, the  pre-requisities  are different.

workingloong:
Fix #217


dlrover/python/master/monitor/speed_monitor.py
```diff
@@ -142,6 +142,9 @@ def reset_running_speed_monitor(self):
         """Reest the speed monitor"""
         self._global_step_records = []
 
+    def achieve_worker_target(self):
```
hxdtest, line 145: `achieve_worker_target` is used to check that all worker are running and starting to get shard from master?
Can it be renamed as `workers_ready`, `all_workers_joined`   or something else so that it is more literal?

workingloong: `all_workers_joined` is good!

hxdtest (approved):
LGTM

PR #241, opened 2023-02-21
Fix bugs in cluster mode testing #1

samplise:


dlrover/examples/iris_job.yaml
```diff
@@ -18,10 +19,12 @@ spec:
             - name: main
               image: easydl/tf-estimator:iris_dnn_v0
               imagePullPolicy: IfNotPresent
+              # yamllint disable
               command:
                 - /bin/bash
                 - -c
-                - "python -m model_zoo.tf_estimator.iris_dnn_elastic"
+                - "python -m model_zoo.tf_estimator.iris_dnn_elastic; sleep 6000"
```
workingloong, line 26: This is an example for users and we'd better not add `sleep 6000`. You can create a new job and set `sleep` in the command to test.

workingloong (approved):

PR #245, opened 2023-02-22
BugFix: Update the iterator of rank_index when updating nodes

workingloong:


dlrover/python/master/node/training_node.py
```diff
@@ -154,10 +154,12 @@ def __init__(
         self._new_node_name_fn = new_node_name_fn
         self._lock = threading.Lock()
         self._node_id_iter = itertools.count(len(self._nodes))
+        self._rank_id_iter = itertools.count(len(self._nodes))
```
hxdtest, line 157: What is the difference between rank_id and node_id?
When will we use rank_id and when will we use node_id?

hxdtest (approved):

PR #248, opened 2023-02-23
[Bug Fix]  Fix autoScale: False type error

hxdtest:
When set 
```
    worker:
      autoScale: False
      template:
```
in the yaml.

spec.get("autoScale", "True") is False and type is bool


workingloong (approved):
LGTM

PR #269, opened 2023-03-02
Optimizer manager implementation

samplise:
Add optimize request processor implementation that process optimize requests from jobs.


workingloong (approved):
LGTM

PR #301, opened 2023-03-13
Fix setting the relaunched number of worker.

workingloong:


dlrover/python/master/node/job_manager.py
```diff
@@ -194,21 +194,21 @@ def _init_training_node_manager(self):
         self._chief_manager = ChiefManager(
             self._job_nodes.get(NodeType.CHIEF, {}),
             self._job_resource,
-            self._ps_relaunch_max_num,
```
hxdtest, line 197: Why not keep _ps_relaunch_max_num？

workingloong: There are `_ps_relaunch_max_num` for the PS and `_relaunch_on_worker_failure` for the worker. It is a bug to set `_ps_relaunch_max_num`  to workers.

workingloong: > Why not keep _ps_relaunch_max_num？

There are `_ps_relaunch_max_num` for the PS and `_relaunch_on_worker_failure` for the worker. It is a bug to set `_ps_relaunch_max_num`  to workers.

hxdtest (approved):

PR #335, opened 2023-03-25
fix dependency conflicts caused by auto best match during "python setup.py install"

Antlera:
python env : conda virtual env python 3.8
During running python setup.py install
image
Because the kubernetes requires a urllib3 verision >=1.24.2, the auto best match version could be 2.0.0a. Here comes the conflict.
image
I modified the requirement.txt to avoid the conflicts.

### DIFF
--- a/dlrover/requirements.txt
+++ b/dlrover/requirements.txt
@@ -1,6 +1,7 @@
 grpcio==1.34.1
 grpcio-tools==1.34.1
-protobuf>=3.12.0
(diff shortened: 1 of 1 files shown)


workingloong (approved):
Excellent!

merlintang:
good try

merlintang:
LGTM

merlintang (approved):

PR #343, opened 2023-03-28
The Tutorial Has Some Minor Bugs

ChenhuiHu:
I found that the elasticjob-name here does not correspond to the previous content, and there should be some minor errors here.


workingloong (approved):
Good!

PR #344, opened 2023-03-28
Environment Test before Start In One Step

ChenhuiHu:
I found that there are some preconditions before downloading and installing this project, which if ignored will cause some inevitable errors later. So I tried to write a document that can use some simple codes to test whether the current environment meets the requirements of this project, thereby reducing the difficulty for users to start this project.


workingloong:
We can move the document into docs/tutorial/check_env.md

merlintang (commented):
good try

merlintang (approved):

PR #347, opened 2023-03-28
Add the yaml to deploy the ElasticJob operator on K8s.

workingloong:


merlintang:
show we use the kustomize ? this would be much easier to maintian.

workingloong:
> show we use the kustomize ? this would be much easier to maintian.

We can get the detail in the Makefile.

https://github.com/intelligent-machine-learning/dlrover/blob/master/dlrover/go/operator/Makefile#L94-L96

PR #353, opened 2023-03-30
Fix the source to install packages.

workingloong:


hxdtest (approved):

docker/Dockerfile
Antlera, line 1: Good to change the source to open source.

PR #363, opened 2023-03-30
Fix some little mistake in docs/deployment/controller.md

ChenhuiHu:
When using the command in the original document, an error that the file path is incorrect will be reported. 
I found that the path in the command is wrong, and it has been modified and tested without any problems.


workingloong (approved):
LGTM

PR #367, opened 2023-03-31
Add set custom reader

hxdtest:


model_zoo/tf_estimator/estimator_executor/conf.py
```diff
@@ -11,9 +11,90 @@
...
+        dataset_name=dataset_name,
+        batch_size=batch_size,
+        num_epochs=num_epochs,
+        dataset_size=dataset_size,
+        num_minibatches_per_shard=num_minibatches_per_shard,
+    )
+    return sharding_client
+
+
+class FakeReader:
```
workingloong, line 40: It is better to define an abstract class as the interface. Users can implement their readers by inheriting the abstract class. For example:

```Python
from abc import ABCMeta, abstractmethod

class ElasticReader(metaclass=ABCMeta):
    def __init__(self, path, num_epochs, batch_size):
         ...
    @abstractmethod
    def count(self):
       pass

    @abstractmethod
    def read(self, start_index, end_index):
          pass

    def iterator(self):
        while True:
            shard = self._data_shard_client.fetch_shard()
            if shard is None:
                break
            for data in self.read(shard.start, shard.end):
                yield data



```

dlrover/trainer/tensorflow/util/reader_util.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@singleton
+class ReaderRegistry:
+    """ """
+
+    def __init__(self):
+        self.reader_registry = {}
+
+    def register_reader(self, reader_name, reader_class):
```
workingloong, line 25: `dataset_name` is better than `reader_name` tell users dlrover trainer use the reader to read which dataset.

dlrover/trainer/tensorflow/util/dataset_util.py
```diff
@@ -43,15 +41,18 @@ def __init__(
...
-            )
-        elif scheme == TFConstants.FAKE_SCHEME():
-            self.reader = FakeReader()  # type: ignore
+        self.initialize_reader(scheme, path)
 
-        if reader_fn is not None:
-            self.reader = reader_fn
+    def initialize_reader(self, scheme, path):
+        reader_type = scheme.split(":")[0]
+        print(reader_type)
```
workingloong, line 48: Remove it or use `log`.

dlrover/trainer/tensorflow/util/dataset_util.py
```diff
@@ -43,15 +41,18 @@ def __init__(
...
+        Reader = reader_registery.get_reader(reader_name=reader_type)
```
workingloong, line 49: Here, users cannot define different readers for different datasets.  It is better to `get_reader` by the dataset name.

dlrover/trainer/tensorflow/reader/base_reader.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class ElasticReader(metaclass=ABCMeta):
+    def __init__(
+        self,
+        path=None,
+    ):
+        self.path = path
+        self.enable_dynamic_sharding = True
+        self.num_minibatches_per_shard = 10
```
workingloong, line 44: Add "_" as prefix to private attributes like `self._path`.

docs/tutorial/add_custom_reader.md
```diff
@@ -0,0 +1,78 @@
...
+```python
+from dlrover.trainer.tensorflow.reader.base_reader import (  # noqa: F401
+    ElasticReader,
+)
+class CustomReader(BaseReader):
+    sepeicif_param = "1"
+
+    def __init__(
+        self,
+        path=None,
```
workingloong, line 22: Edit the document according to the new definition of `ElasticTrainer`.

workingloong (approved):
LGTM

Issue #370, opened 2023-04-01
After pytorch-mnist starts, elasticjob pod always error.


ChenhuiHu:
The referenced document is dlrover/docs/deployment/controller.md . After starting the task with `kubectl -n dlrover apply -f dlrover/examples/torch_mnist_master_backend_job.yaml` every time, the elasticjob pod status will show Error.
[2023_4_1（torch-minst启动后elasticjob总是error）En.pdf](https://github.com/intelligent-machine-learning/dlrover/files/11128290/2023_4_1.torch-minst.elasticjob.error.En.pdf)

ChenhuiHu:
Problem solved! #371 fixes what caused the bug.

PR #371, opened 2023-04-01
Fix role creating in dlrover namespace.

Antlera:
User may create role in default namespace by following the outdated doc.


Antlera:
Fix #370

workingloong (approved):
LGTM

PR #380, opened 2023-04-05
Set the max number of shards for workers who can exits together for sync training .

workingloong:


dlrover/python/elastic_agent/sharding/client.py
```diff
@@ -49,7 +52,7 @@ def __init__(
         dataset_size,
         shuffle=False,
         task_type=elastic_training_pb2.TRAINING,
-        num_minibatches_per_shard=0,
+        num_minibatches_per_shard=10,
```
merlintang, line 55: 10 is a magic number?

hxdtest (approved):
LGTM

PR #385, opened 2023-04-09
Update developer_guide.md

ChenhuiHu:
Fixed some bugs in this documentation, and supplemented the section on obtaining more GPU resources.

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -57,19 +57,59 @@ and [NVIDIA Container Toolkit](https://github.com/NVIDIA/nvidia-docker)
 
 It is highly recommended to have more than one GPU resource in your workspace.
 
-However, there is still a workaround to divide your singie GPU resource into multiple ones.
+However, there is still a workaround to divide your single GPU resource into multiple ones.
 
 For this, enable [shared-access-to-gpus with CUDA Time-Slicing](https://github.com/NVIDIA/k8s-device-plugin#shared-access-to-gpus-with-cuda-time-slicing) to get more GPU resources.
 
-Check the doc and modify your ``nvidia-k8s-device-plugin`` or simply update the plugin by ``helm`` with the command
+Check the doc and modify your ``nvidia-k8s-device-plugin`` or simply update the plugin by ``helm`` with the command (make sure start you minikube or something else before these steps)
 
-```bash
-$ helm upgrade -i nvdp nvdp/nvidia-device-plugin \
-    --version=0.13.0 \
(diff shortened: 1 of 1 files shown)


docs/developer_guide.md
```diff
@@ -57,19 +57,59 @@ and [NVIDIA Container Toolkit](https://github.com/NVIDIA/nvidia-docker)
 
 It is highly recommended to have more than one GPU resource in your workspace.
 
-However, there is still a workaround to divide your singie GPU resource into multiple ones.
+However, there is still a workaround to divide your single GPU resource into multiple ones.
 
 For this, enable [shared-access-to-gpus with CUDA Time-Slicing](https://github.com/NVIDIA/k8s-device-plugin#shared-access-to-gpus-with-cuda-time-slicing) to get more GPU resources.
 
-Check the doc and modify your ``nvidia-k8s-device-plugin`` or simply update the plugin by ``helm`` with the command
+Check the doc and modify your ``nvidia-k8s-device-plugin`` or simply update the plugin by ``helm`` with the command (make sure start you minikube or something else before these steps)
```
workingloong, line 64: It is better to move the content to install GPU for minikube into a single document and add a link to the document.

ChenhuiHu: This part of the documentation is for testing tasks on a single machine. The part I modified is mainly to split one GPU of the stand-alone machine into two for use, and the original command will have an error message when it is used directly, which may cause some confusion. 

image

image

So I added two ways of install and upgrade, and made some changes in the original command to avoid unexpected error. At the same time, because this part is not very lengthy, I wonder if adding an additional document will cause confusion in reading?

workingloong: > This part of the documentation is for testing tasks on a single machine. The part I modified is mainly to split one GPU of the stand-alone machine into two for use, and the original command will have an error message when it is used directly, which may cause some confusion.
> 
> image
> 
> image
> 
> So I added two ways of install and upgrade, and made some changes in the original command to avoid unexpected error. At the same time, because this part is not very lengthy, I wonder if adding an additional document will cause confusion in reading?

No need to worry about it. In the document, we also only give the link to install the minkube not the detailed steps. The document to set up the minikube with GPU is not only for DLRover. What's more, the document is introduced to the developer of DLRover and we should reduce the content that is irrelative to DLRover.

ChenhuiHu: OK, got it. Thank you for your reply!

workingloong (approved):

PR #387, opened 2023-04-10
A doc to develop a PyTorch model using elastic training of DLRover.

workingloong:


docs/tutorial/pytorch_training.md
```diff
@@ -0,0 +1,175 @@
...
+with elasticity using DLRover. Users only need to make some simple changes
+of native PyTorch training codes. We have provided the
+[CNN example](../../model_zoo/pytorch/mnist_cnn.py) to show how to
+train a CNN model with the MNIST dataset.
+
+## Develop a Torch Model with DLRover. 
+
+### Setup the Environment Using ElasticTrainer
+
+Users need to set up the environment by `ElasticTrainer`. The `ElasticTrainer`
```
samplise, line 13: `by` -> `through`

docs/tutorial/pytorch_training.md
```diff
@@ -0,0 +1,175 @@
...
+will set the IP of rank-0 into `MASTER_ADDR`, because the node with rank-0
```
samplise, line 14: `because` -> `and`, `will` -> `will not` ?

docs/tutorial/pytorch_training.md
```diff
@@ -0,0 +1,175 @@
...
+
+```python
+from dlrover.trainer.torch.elastic import ElasticTrainer
+
+ElasticTrainer.setup()
+```
+
+### Develop the ElasticDataset.
+
+Firstly, users need to write the path of the sample into a `Text` file.
```
samplise, line 25: `Firstly` -> `At first`

docs/tutorial/pytorch_training.md
```diff
@@ -0,0 +1,175 @@
...
+/data/mnist_png/training/9/374.png,9
+/data/mnist_png/training/1/29669.png,1
+/data/mnist_png/training/1/19782.png,1
+/data/mnist_png/training/1/42786.png,1
+/data/mnist_png/training/8/41017.png,8
+/data/mnist_png/training/8/13037.png,8
+/data/mnist_png/training/8/7101.png,8
+```
+
+Then, we can develop a dataset to support elastic training
```
samplise, line 42: delete `can`

docs/tutorial/pytorch_training.md
```diff
@@ -0,0 +1,175 @@
...
+        """
+        pass
+```
+
+### Wrap the Training Step using ElasticTrainer
+
+To keep the total batch size fixed during elastic training,
+users need to create an `ElasticTrainer` to wrap the model, optimizer
+and scheduler. `ElasticTrainer` can keep the total batch size
+fixed by accumulating gradients if the number of worker decreases.
```
samplise, line 86: It'd better also explain how to fix the batch size when the number of workers increases here.

PR #422, opened 2023-06-01
make distributed work for atorch.

nash635:
make distributed work for atorch.

### DIFF
--- a/atorch/setup.py
+++ b/atorch/setup.py
@@ -20,6 +20,7 @@ setup(
     # 0.4.2,require python3.7
     install_requires=required_deps,
     package_data={"": ["*.so"]},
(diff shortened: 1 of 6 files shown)


atorch/setup.py
```diff
@@ -16,7 +16,7 @@ def fetch_requirements(path):
     description="A pytorch extension for efficient deep learning.",
     author="Ant Group",
     python_requires=">=3.5",
-    packages=find_packages(exclude=["*test*"]),
```
workingloong, line 19: Why do not we exclude tests?

nash635: used for test, reverted.

workingloong (approved):
LGTM

PR #441, opened 2023-06-13
add auto_accelerate API

skydoorkai:
Add auto_accelerate API with some basic optimization methods. This will support load_strategy with these methods.

This is the first step in adding auto_accelerate.

TODO:
1. add AccelerationEngine to support optimization strategy auto search.
2. add tensor parallel/pipe parallel and other methods.

### DIFF
--- a/.github/actions/atorch-python-test/action.yml
+++ b/.github/actions/atorch-python-test/action.yml
@@ -8,6 +8,7 @@ runs:
     - "/bin/bash"
     - "-c"
     - "pip install pytest fairscale==0.4.1\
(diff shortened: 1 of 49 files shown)


.github/actions/atorch-python-test/action.yml
```diff
@@ -7,6 +7,8 @@ runs:
   args:
     - "/bin/bash"
     - "-c"
-    - "pip install pytest \
+    - "pip install pytest fairscale==0.4.1\
```
nash635, line 10: fairscale==0.4.1 already included in requirments.txt, rebuild docker image will include it.

nash635 (approved):

PR #458, opened 2023-06-27
Rendezvous manager to assign the rank to nodes.

### LINKED ISSUE #450: Rendezvous is closed if some agent fail.
We found some drawbacks when we use TorchElastic to train LLM model on the k8s cluster.
- The agent of TorchElastic will close rendezvous when it exits. The agent of the remaining running Pods can not join the next round rendezvous because it is closed.
- The rendezvous state is stored in a GPU node when rendezvous backend is c10d. All agents will fails if the node breakdowns.
- The rank of nodes is random after a round rendezvous is completed. Users may has trouble to find the rank-0 which has some training logs.

workingloong:
Fix #450


dlrover/python/master/elastic_training/rdzv_manager.py
```diff
@@ -0,0 +1,158 @@
...
+            if self._rdzv_nodes:
+                return self._rdzv_nodes
+            if len(self._waiting_nodes) == self._rdzv_params.max_nodes:
+                rdzv_completed = True
+            else:
+                waiting_num = len(self._waiting_nodes)
+                alive_num = len(self._alive_nodes)
+                waiting_time = time.time() - self._lastcall_time
+                rdzv_completed = (
+                    waiting_num == self._rdzv_params.min_nodes
```
workingloong, line 117: >=

hxdtest (approved):
LGTM

Issue #384, opened 2023-04-08
403 Forbidden when submitting a job.


Antlera:
```bash
kubectl -n dlrover apply -f dlrover/examples/torch_mnist_master_backend_job.yaml
```
```bash
kubectl logs elasticjob-torch-mnist-dlrover-master 
```
```bash
[2023-04-08 09:34:59,828] [INFO] [global_context.py:108:print_config] DLRover global context = {'train_speed_record_num': 50, 'seconds_to_autoscale_worker': 90, 'step_to_adjust_worker': 200, 'optimize_worker_cpu_threshold': 20, 'seconds_for_stable_worker_count': 60, 'seconds_interval_to_optimize': 300, 'factor_to_cut_pending_cpu': 2, 'seconds_to_wait_pending_pod': 900, 'seconds_huge_training_threshold': 1800, 'sample_count_to_adjust_worker': 5, 'seconds_interval_to_change_ps': 3600, 'auto_worker_enabled': False, 'auto_ps_enabled': False, 'is_tfv1_ps': False}
E0408 09:35:00.456777939      11 ev_epollex_linux.cc:516]    Error shutting down fd 7. errno: 9
[2023-04-08 09:35:04,854] [WARNING] [client.py:276:build_brain_client] The GRPC service of brain is not available.
[2023-04-08 09:35:04,854] [WARNING] [client.py:88:__init__] Cannot initialize brain channel
[2023-04-08 09:35:05,240] [INFO] [args.py:65:print_args] port = 50001
[2023-04-08 09:35:05,240] [INFO] [args.py:65:print_args] job_name = torch-mnist
[2023-04-08 09:35:05,240] [INFO] [args.py:65:print_args] namespace = dlrover
[2023-04-08 09:35:05,240] [INFO] [args.py:65:print_args] platform = k8s
[2023-04-08 09:35:05,240] [INFO] [factory.py:31:new_job_args] New k8s JobParameters
[2023-04-08 09:35:05,241] [INFO] [kubernetes.py:67:__init__] Load the incluster config.
[2023-04-08 09:35:05,246] [WARNING] [kubernetes.py:146:get_custom_resource] Exception when getting custom object: (403)
Reason: Forbidden
[...]

PR #493, opened 2023-07-11
Integrate default-role.yaml into Kustomization.

Antlera:
Fix possible permissions issues when deploying the job controller (e.g. #384).


dlrover/go/operator/config/manifests/bases/default-role.yaml
```diff
@@ -0,0 +1,46 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
workingloong, line 2: Remove the replicated file https://github.com/intelligent-machine-learning/dlrover/blob/master/dlrover/go/operator/config/rbac/default_role.yaml

workingloong (approved):
LGTM

PR #515, opened 2023-07-20
Fix files for local debugging.

Antlera:
Fix file naming and mirror pulling policy for local debugging.
Fix docker tag to prevent confusion with controller image.


workingloong (approved):
LGTM

PR #529, opened 2023-07-26
Fix: Fix cuda device settings.

Antlera:
Fix device error possible when cuda driver fails and cuda device is not available.


workingloong (approved):
LGTM

Issue #538, opened 2023-07-27
Add parameter descriptions or code hints for nanogpt example.


Antlera:
For [nanogpt_train_script](https://github.com/intelligent-machine-learning/dlrover/blob/master/model_zoo/pytorch/nanogpt/train.py)
Parameter descriptions and possible values need to be added to help users better understand code functionality.
For example: 
User may pass 'gpu' as a argument, somehow expected device is in the form of "cuda[0-9]".
```python
    # System settings
    parser.add_argument("--device", type=str, default="cpu", required=False)
    parser.add_argument("--compile", type=str, default="False", required=False)
```

PR #545, opened 2023-07-28
[#538]Add device_type function and warning messages

xin-chen42:
To improve the extensibility of parameters, this PR adds the device_type function that can recognize various input device types such as "gpu", "GPU0", "GPU1", "cuda01", etc., and automatically uses the "cuda" device. If the input device type is not recognized, it defaults to using the "cpu" device.
This change provides a better user experience.

### DIFF
--- a/model_zoo/pytorch/nanogpt/train.py
+++ b/model_zoo/pytorch/nanogpt/train.py
@@ -14,7 +14,6 @@
 
 import argparse
 import contextlib
-import logging
 import math
 import os
 import pickle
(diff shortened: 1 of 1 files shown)


model_zoo/pytorch/nanogpt/train.py
```diff
@@ -319,7 +319,24 @@ def train():
         if iter_num > max_iters:
             break
 
-
+def device_type(string):
+    lower_string = string.lower()
+    if "gpu" in lower_string or "cuda" in lower_string:
+        if lower_string != "cuda":
+            print(
```
workingloong, line 326: We can use logging.warning to print the warning log.

workingloong (approved):
LGTM

workingloong:
A hint: It is better to rename the original branch `xin-chen42:master` to a meaningful name like `xin-chen43:add_device_type_to_nanogpt`.

PR #555, opened 2023-07-31
Reduce the resource if the pending time is too long.

workingloong:

If the pending time is too long because of not enough resources,  the master need to reduce the node resource.

### DIFF
--- a/dlrover/python/master/node/training_node.py
+++ b/dlrover/python/master/node/training_node.py
@@ -194,7 +194,6 @@ class TrainingNodeManager(object):
             new_id = next(self._node_id_iter)
             relaunch_node = node.get_relaunch_node_info(new_id)
             self._nodes[new_id] = relaunch_node
-            node.relaunchable = False
         logger.info("Relaunch node %s to %s", node.name, new_id)
         plan.launch_nodes.append(
             Node(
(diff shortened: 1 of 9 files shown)


dlrover/python/master/node/training_node.py
```diff
@@ -127,9 +128,23 @@ def cut_timeout_pending_node_cpu(node: Node):
             _dlrover_context.seconds_to_wait_pending_pod,
             new_cpu,
         )
-        return True
-    else:
-        return False
+        reduced = True
+    original_memory = node.config_resource.memory
+    new_memory = math.ceil(
+        original_memory / _dlrover_context.factor_to_cut_pending_cpu
```
samplise, line 134: Should it be `factor_to_cut_pending_mem`? Also, I think we'd better set a threshold, e.g., 32g. When the original memory is more than the threshold, we conduct resource reduction.

workingloong: Now, we reduce the memory only when `original memory/ factor_to_cut_pending_mem > NodeResourceLimit.MIN_MEMORY` where NodeResourceLimit.MIN_MEMORY= 6GB.

samplise (approved):

PR #557, opened 2023-07-31
[#538]Add device_type function and warning messages

xin-chen42:

This pull request proposes a modification to the device_type function in the codebase. The changes involve determining the device type based on the input string, which is used to specify the device type (e.g., "gpu" , "Gpu01"or "cuda" for GPU, and "cpu" for CPU).

The purpose behind these changes is to improve the device type determination logic in the device_type function. 

Yes, this PR introduces user-facing changes in the form of log messages. If the input string indicates a GPU device (e.g., "gpu" or "cuda"), the function now handles the variations in user input and provides a more informative message. If the input string is not recognized, the function also gives feedback about the unrecognized argument and defaults to using the CPU device.

Use different parameter configurations for the 'device' argument and check the relevant logs.

### DIFF
--- a/model_zoo/pytorch/nanogpt/train.py
+++ b/model_zoo/pytorch/nanogpt/train.py
@@ -14,7 +14,6 @@
 
 import argparse
 import contextlib
-import logging
 import math
 import os
 import pickle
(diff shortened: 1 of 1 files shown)


Antlera (approved):
LGTM

PR #587, opened 2023-08-07
Support infinite epoch if the num_epoch is less than 0.

workingloong:

Support infinite epoch if the num_epoch <= 0

Users may need an infinite  number of epoch and stop training by the max step.


hxdtest (approved):
LGTM

nash635 (approved):

PR #592, opened 2023-08-08
Completed the parameters passed to the get_lr() and Replaced Adam with AdamW

xin-chen42:

Completed the parameters passed to the get_lr() and Replaced Adam with AdamW

The get_lr() function is not being passed the correct arguments. 
AdamW is better than Adam.

None.

Use "--decay_lr" parameters


workingloong (approved):
LGTM

PR #613, opened 2023-08-14
3/3 feat(master.job_manager) add updating GPU statistic to job manager.

Antlera:

Add updating GPU statistic to job manager.

We use GPU statistics collected by the monitor to update node resources awaiting a later optimizing plan.

NO.

Tests are performed using the included unit test code.


dlrover/python/master/node/local_job_manager.py
```diff
@@ -60,9 +60,11 @@ def early_stop(self):
     def add_node_event_callback(self, node_event_callback):
         pass
 
-    def update_node_resource_usage(self, node_type, node_id, cpu, memory):
+    def update_node_resource_usage(
```
workingloong, line 63: Add a docstring to explain what is the gpu_stats. You can give an example of gpu_stats in the docstring.

dlrover/python/scheduler/kubernetes.py
```diff
@@ -351,13 +351,16 @@ def initilize(self):
...
             gpu_num = 0
+            gpu_stats = []
             for k, v in requests.items():
                 if "nvidia.com" in k:
                     gpu_type = k
                     gpu_num = int(v)
             group_resource = NodeGroupResource(
                 num,
-                NodeResource(cpu, memory, gpu_type, gpu_num, priority),
+                NodeResource(
```
workingloong, line 361: Why do we need to modify these codes?

Antlera: Now the `NodeResource` is initialized with gpu_stats. We need to add an empty list to initialize the `NodeResource`.

Antlera: But it's also strange here, ideally kubernetes.py should only handle resource configuration requests from yaml. However `gpu_stats` is not part of that.

dlrover/python/master/node/local_job_manager.py
Antlera: Update the docstring in `NodeResource` like this：
```
class NodeResource(JsonSerializable):
    """NodeResource records a resource of a Node.
    Attributes:
        cpu: float, CPU cores.
        memory: float, memory MB.
        gpu_type: str, the type of GPU.
        gpu_num: int,
        gpu_stats: list of GPUMetric, a list of GPUMetric dataclass objects,
        each containing GPU statistics.
            - index (int): The index of the GPU.
            - total_memory_mb (int): Total GPU memory in megabytes.
            - used_memory_mb (int): Used GPU memory in megabytes.
            - gpu_utilization (float): GPU utilization in percentage (0.0 to
              100.0).
        image: the image name of the node.
        priority: the priority classs of the node.
    Example:
        To update the resource usage of a worker node with ID 123 and provide
        GPU stats:
        >>> gpu_stats = [GPUMetric(index=0, total_memory_mb=8192,
        used_memory_mb=2048, gpu_utilization=80.0)]
        >>> cluster_manager.update_node_resource_usage('worker', 123, 75.0,
        50.0, gpu_stats)
    """
```
or
Add new docstring to `update_node_resource_usage`?

workingloong: Updation of `NodeResource`.

workingloong (approved):
LGTM

PR #616, opened 2023-08-15
Refactor: Add start and stop method to ResourceMonitor.

workingloong:

Add start and stop method to ResourceMonitor.

We can control the time to start and stop monitoring the resource.

No.

No.

### DIFF
--- a/dlrover/python/elastic_agent/monitor/resource.py
+++ b/dlrover/python/elastic_agent/monitor/resource.py
@@ -97,8 +97,8 @@ class ResourceMonitor(object):
 
     def start(self):
         if (
-            os.getenv(NodeEnv.DLROVER_MASTER_ADDR, "")
-            and os.getenv(NodeEnv.AUTO_MONITOR_WORKLOAD, "") == "true"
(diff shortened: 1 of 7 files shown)


dlrover/python/elastic_agent/monitor/resource.py
```diff
@@ -92,32 +92,45 @@ def __init__(self):
...
-                    daemon=True,
-                )
-                thread.start()
-                if thread.is_alive():
-                    logger.info("Resource Monitor initialized successfully")
-            except Exception as e:
-                logger.error(
-                    f"Failed to start the monitor resource thread. Error: {e}"
-                )
+            return
```
Antlera, line 103: Why we should return here when `MASTER_ADDR` & `AUTO_MONITOR` is `True`?

workingloong: This is a bug and I have fixed it.

Antlera:
LGTM

Antlera (approved):
LGTM

PR #667, opened 2023-09-06
Set the job name and uid into the worker env.

workingloong:

Set the job name and uid into the worker env.### Why are the changes needed?

No.

UT.


Major-333 (approved):
LGTM

PR #669, opened 2023-09-07
Bugfix: all agent can update rendezvous params.

workingloong:

Remove the limit where only the rank0 worker report rdzv params to the master.

The worker whose rank is not 0 may start before the worker-0.

No.

UT.


hxdtest (approved):
LGTM

PR #682, opened 2023-09-11
2/3 feature(auto_tuning) Update servicer for paral_config.

Antlera:

Update servicer for paral_config.

This is one component of the auto_tuning feature.

No.

UT.


dlrover/python/elastic_agent/config/paral_config_tuner.py
```diff
@@ -0,0 +1,68 @@
...
+class ParalConfigTuner(object):
+    def __init__(self):
+        """
+        Parallelism config tuner for updating parallelism config file.
+        """
+        self._master_client = GlobalMasterClient.MASTER_CLIENT
+        self.config_dir = os.path.dirname(
+            os.environ[ConfigPath.ENV_PARAL_CONFIG]
+        )
+        self.config_path = os.environ[ConfigPath.ENV_PARAL_CONFIG]
```
workingloong, line 35:
```
self.config_path = os.getenv(ConfigPath.ENV_PARAL_CONFIG, "")
```
We can avoid the error if the ConfigPath.ENV_PARAL_CONFIG is not in `os.environ`.

dlrover/python/master/node/dist_job_manager.py
```diff
@@ -28,8 +28,12 @@
     NodeType,
 )
 from dlrover.python.common.global_context import Context
+from dlrover.python.common.grpc import ParallelConfig
 from dlrover.python.common.log import default_logger as logger
 from dlrover.python.common.node import Node, NodeGroupResource
+from dlrover.python.master.hyperparams.local_strategy_generator import (
```
workingloong, line 34: Do you miss the `local_strategy_generator`?
It is better to rename `local_strategy_generator` to `strategy_generator` because the local job or distributed job
can both use it.

PR #683, opened 2023-09-12
Fix: check the config file to update batch size after a step.

workingloong:

ElasticTrainer checks the config file to update batch size after a step.

We can dynamically update the batch size to improve throughput.

Users need to initialize `ElasticTrainer` withe the dataloader.

UT


merlintang (approved):
do we need to change to ElasticDataLoader to support the Elastic ?

PR #688, opened 2023-09-13
Doc: Design to detect straggler before training.

workingloong:

A design document to introduce how to detect the straggler before training.

It is difficult for users to search the straggler.

No.

It is just a document.


merlintang (commented):
I have one question here:

does this process run as the schedule job like each hour or daily? if one straggler node is found, does this influence the following scheduling procedure?

workingloong:
> I have one question here:
> 
> does this process run as the schedule job like each hour or daily? if one straggler node is found, does this influence the following scheduling procedure?

Now, the process is a tool for users to detect whether there is a straggler in a job. Sometime, users may suspect there is a straggler in the nodes of a job and need to manually check all nodes by binary search. I hope users can run a simple task in a job to quickly check whether there is a straggler.

If one straggler is found, the process needs to notify users which node is the straggler. In the future, we can intelligently assign the light training task to the straggler. For example, we can turn down the mini-batch size of the straggler.

merlintang:
LGTM

merlintang (approved):

PR #699, opened 2023-09-14
Check whether the config file is empty.

workingloong:


Antlera:
A very inspiring corner case checking! This is was a valuable learning experience for me.

PR #701, opened 2023-09-15
Feat: The script to check whether the GPU is available.

workingloong:

Add scripts to check whether the GPU is available.

Users may need to check whether the GPU is available before the training starts.

No.

`python -m dlrover.trainer.check.gpu_check`  on a GPU node.


cailun01 (approved):
LGTM

PR #718, opened 2023-09-21
Detect the straggler by running a simple task.

### LINKED ISSUE #678: Automatically detect the straggler.
In the collective communication job, the straggler will significantly affect the performance of training. If there are many nodes in a cluster, it is difficult to detect which node is the straggler. The reason of the straggler may be slow RDMA or nvlink or unknown hardware failure.

workingloong:

The agent runs a simple task to detect the straggler.

It is difficult for users to detect the straggler if the number of nodes in a job is large.

Fix #678 

`dlrover-run --network-check`

UT and create a job to run `dlrover-run --network-check`.
image


hxdtest (approved):
LGTM

PR #758, opened 2023-10-11
remove deprecated optim, fix optim ut

ZhongYingMatrix:
Remove the outdated optimizer, unified in `optimizers` folder. Fix ut `optim_test.py` since `load_state_dict` alteration in https://github.com/pytorch/pytorch/pull/106082


skydoorkai:
Only delete apex-based optim, and keep bf16optimizer, wsam, etc.

skydoorkai (approved):
lgtm now.

PR #766, opened 2023-10-17
Fix AGD error

yzlnew:

Fix AGD.

`self.weight_decouple` and `self.fixed_decay` is not valid in previous implementation.

No.

On internal training scripts, not public yet.


workingloong (approved):
LGTM

PR #793, opened 2023-11-01
Support configuring the range of PyTorch distributed master port for TCPStore

workingloong:

Add an argument to set the range of PyTorch distributed master port  for TCPStore.

The random port may be grabbed by other processes when the network of Pod is the host.

`dlrover-run --master_port_range 10000:20000`

`kubectl -n dlrover apply -f examples/pytorch/mnist/elastic_job.yaml`


hxdtest (approved):
LGTM

PR #808, opened 2023-11-09
Doc: Design to asynchronously checkpoint model states into the storage.

workingloong:

A design document to asynchronously checkpoint model states into the storage.

Asynchronous checkpointing can significantly reduce the time to block the training when saving model states into the storage.

No.

Not need.

### DIFF
--- a/docs/design/async-checkpoint.md
+++ b/docs/design/async-checkpoint.md
@@ -9,7 +9,7 @@ It is very difficult to keep training a fundational model during a lone time
 without any interruption. Hareware failure, OS failure or network breakdwon
 may oftern happen during the training. Users usually periodically checkpoint
 the model and optimizer states to the storage to resume the training
-after an interruption. Now, It takes a few minutes to use  `torch.save` to 
+after an interruption. Now, It takes a few minutes to use  `torch.save` to
 checkpoint a large fundational model. Frequent checkpointing may waste
 many hours of the accelerator (GPU, NPU, TPU).
 Because, the training must stop until the checkpointing is completed.
@@ -20,7 +20,7 @@ when we train a fundation model using thouands of accelerators.
 
 ## Target
 
-- Shorten the checkpointing time overhead to block the training for 
+- Reduce the checkpointing time overhead to block the training for
  DDP, FSDP, DeepSpeed or Megatron.
 - Speed up loading the checkpoint file when the training restarts.
 - Automatically configure the checkpoint period to improve the efficient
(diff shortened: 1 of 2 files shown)


docs/design/async-checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+## Background
+
+It is very difficult to keep training a fundational model during a lone time
+without any interruption. Hareware failure, OS failure or network breakdwon
+may oftern happen during the training. Users usually periodically checkpoint
+the model and optimizer states to the storage to resume the training
+after an interruption. Now, It takes a few minutes to use  `torch.save` to
+checkpoint a large fundational model. Frequent checkpointing may waste
+many hours of the accelerator (GPU, NPU, TPU).
+Because, the training must stop until the checkpointing is completed.
```
samplise, line 15: `accelerator (GPU, NPU, TPU). Because, the training` - > `accelerator (GPU, NPU, TPU) since the training`

docs/design/async-checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+On the other hand, too long checkpionting period will waste many training steps
+when resuming the training from the latest checkpoint file. Shortening the checkpinting
```
samplise, line 17: Here do you mean the large size of checkpoints require long time period to resume?

workingloong: It is not. For example, we save a checkpoint per 24h and the training fails at 23h. The training resumes the model from the 23 hours old checkpoint. The 23 hours are wasted during the latest checkpointing and the interruption.

samplise: Ok. I think it should be more clear like `On the other hand, infrequent checkpoints could reduce the overhead but take the risk of wasting more training steps when ...`

docs/design/async-checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+
+It is very difficult to keep training a fundational model during a long time period
+without any interruption. Hareware failure, OS failure or network breakdwon
+may oftern happen during the training. Users usually periodically checkpoint
+the model and optimizer states to the storage and resume the training
+after an interruption. Now, It takes a few minutes to use  `torch.save` to
+checkpoint a large fundational model. Frequent checkpointing may waste
+many hours of the accelerator (GPU, NPU, TPU) since the training must stop
+until the checkpointing is completed. On the other hand, infrequent checkpoints
+ould reduce the overhead but take the risk of wasting more training steps
```
samplise, line 16: `ould` -> `could`

samplise (approved):

PR #816, opened 2023-11-13
fix bool error when tp size 1

ZhongYingMatrix:
Fix `RuntimeError: Boolean value of Tensor with more than one value is ambiguous` when tp size 1


skydoorkai (approved):
lgtm

workingloong (approved):
LGTM

PR #830, opened 2023-11-19
Clean up the temp file to test ray job args.

workingloong:


dlrover/python/tests/test_ray_job_args.py
```diff
@@ -25,6 +25,9 @@
 
 
 class RayJobArgsTest(unittest.TestCase):
+    def tearDown(self):
+        os.remove("test.json")
+
     def test_initialize_params(self):
         file = "data/demo.yaml"
```
hxdtest, line 32: it would be better to use `tempfile` to manage temporary file.

workingloong: Can you fix it later? I don't know what is the usage of `test.json` in the UT.

hxdtest (approved):

PR #832, opened 2023-11-20
add flash attention gpu ut

mars1248:
add flash attention gpu ut


workingloong:
We do not have any GPU to execute the UT. How do we validate it?

workingloong (approved):
LGTM

PR #852, opened 2023-11-24
Add design doc to implement the async checkpointing with agent process.

liyzcj:
Design to implement async checkpoint with agent process and Last words Feature


docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+thread per node to save the checkpointing state into the storage, we design three classes to
+implement the checkpointing process.
+
+<div align="center">
+<img src="../figures/async-ckpt-classes.jpg" alt="Async Checkpoint Classes" width="1000">
+</div>
+
+- **GlobalCkptManager**
+    - Only one instance runs in the Rank 0 Agent process.
+    - The class is responsible for keeping the consistency of the checkpointing state in the shared
```
workingloong, line 148: Why and how does the manager keep the consistency of the checkpointing state in the shared memory?

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+      memory and the storage.
+    - Get the read lock of shared memory and notify all agents to save the checkpointing state
```
workingloong, line 150: It is not necessary to notify all agents to save. Another way is, the`TrainCkptManager` of local rank 0 on each node notifies its agent to save the state into the storage when the iteration step is satisfied. `TrainCkptManager` on all nodes can synchronously notify its agent in a synchronous SGD job.

workingloong: Can we merge `GlobalCkptManager` and `AgentCkptManger` into a class?

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+      into the storage, update the checkpointing tracker file in the storage.
+- **AgentCkptManger**
+    - One instance runs in each agent process.
```
workingloong, line 153: It also need to allocate the shared memory blocks for each rank according to the size of checkpointing state dict.

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+    - Responsible for saving the checkpointing state from shared memory into the storage.
```
workingloong, line 154: Be responsible for

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+    - Notify the GlobalCkptManager when the saving procedure is completed.
```
workingloong, line 155: Notify -> Notifies.

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+- **TrainCkptManger**
+    - One instance runs in each training process.
+    - Responsible for saving the checkpointing state from GPU to shared memory.
```
workingloong, line 158: Be responsible for

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+    - Notify the GlobalCkptManager when we need to save the checkpointing state into the storage.
```
workingloong, line 159: Notify -> Notifies.

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+
+The agent and training process need to do the following steps:
+
+- Init Process - first checkpoint to memory
+
+    - The agent monitors the training process to create the mata of model and
+      optimizer state dict.
+    - The training process notifies GlobalCkptManager the checkpointing meta when the
+      training process firstly saves the state dict into the memory.
+    - GlobalCkptManager acquire the lock of the shared memory and notify all agents to
```
workingloong, line 175: acquire -> acquires.

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+    - The TrainCkptManger acquires the lock of the shared memory and copies the state dict
+      from GPU to the shared memory.
+
+- Training Loop
+
+    - TrainCkptManager acquires the lock of the shared memory and copies the state dict
+      from GPU to the shared memory.
+    - TrainCkptManager notifies GlobalCkptManager to save the checkpointing state into the
+      storage if the iteration step is multiple of the save_interval.
+    - GlobalCkptManager acquires the lock of the shared memory and notifies all agents to
```
workingloong, line 186: It is not necessary to notify all agents. The agent saves its state dict to the storage when the training process notify it to save.

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+      save the checkpointing state into the storage.
+    - AgentCkptManger write the checkpointing state into the storage and notifies
```
workingloong, line 188: write -> writes.
It is not necessary to notify.

docs/design/async-checkpoint.md
```diff
@@ -120,6 +124,93 @@ training process to resume the training, the training process can load the check
...
+</div>
+
+### Last Words when the Training Process Fails
+
+When any of the training processes fails, or the agent is killed by SIGTERM, we can automatically
+save the latest checkpointing state into the storage.
+
+### Consistency of the Checkpointing State
+
+For the shared memory checkpointing consistency and correctness, we use a distributed lock
```
workingloong, line 206: A shared lock is more accurate than a distributed lock.

docs/design/async-checkpoint.md
liyzcj: for simplicity and stability, we use a polling instead of notification.

docs/design/async-checkpoint.md
liyzcj: for simplicity and stability, we use a polling instead of notification, and use one manager with a if statement to polling the checkpoint state.

docs/design/async-checkpoint.md
```diff
@@ -166,29 +162,17 @@ implement the checkpointing process.
...
-      save the checkpointing state into the storage.
-    - AgentCkptManger write the checkpointing state into the storage and notifies
-      GlobalCkptManager when finished.
-    - GlobalCkptManager commits the checkpointing and modify tracker file after all agents
-      finish the writing.
+1. The agent monitors the training process to create the mata of model and
+  optimizer state dict.
+2. TrainCkptManager acquires the shared lock and update the meta of model and optimizer
+  state dict.
+3. TrainCkptManager copy the state dict from GPU to the shared memory.
```
workingloong, line 169: copy -> copies

PR #875, opened 2023-12-07
Refactor CheckpointSaver to support Megatron-LLM.

workingloong:

Refactor CheckpointSaver to support Megatron-LLM.

There are many repeated codes in `NoShardingCheckpointSaver` and `ShardingCheckpointSaver`.

Yes.

UT.


dlrover/python/elastic_agent/torch/ckpt_saver.py
```diff
@@ -722,91 +797,139 @@ def _commit_checkpoint(
...
 
-        step = self._shm_handlers[0].get_iteration_step()
-        # we need to release the locks before save to storage
+        queue.put(class_meta)
+        queue.unlink()
 
-        for lock in self._shm_lock:
-            lock.release()
+    @timer
+    def save_to_memory(self, step, state_dict, path=""):
```
liyzcj, line 953: we may need a block parameter  here, when user is calling save_to_storage, we should update memory synchronously, or we may loss the step user wanted.

dlrover/python/elastic_agent/torch/ckpt_saver.py
```diff
@@ -722,91 +797,139 @@ def _commit_checkpoint(
...
 
-        if not self._acquire_all_locks():
-            # The training process does not release the lock because it fails
-            # when writing the state dict into the shared memory. The shared
-            # memory may be dirty and the saver cannot save it to the storage.
+    def _notify_agent_to_create_saver(self):
+        if self._local_rank != 0:
             return
+        if self._restart_count > 0:
+            # Only local rank 0 notify to initialize the saver in
```
liyzcj, line 929: since the saver only initialized once, if an agent start with a new global_shard_num in elastic training, we can't reset this parameter, this may cause problem when saving to storage.

dlrover/python/elastic_agent/torch/ckpt_saver.py
workingloong: save_to_storage will firstly call save_to_memory and then notify an event to the agent to save.

dlrover/python/elastic_agent/torch/ckpt_saver.py
liyzcj: we may change the lifecycle of Saver，one instance for each training attempt。

dlrover/python/elastic_agent/torch/ckpt_saver.py
liyzcj: yes，but if the agent is writing to storage now, the save_to_memory can't acquire the lock, then the memory sync will be skipped

workingloong: It is reasonable to skip `save_to_memory` if the agent is writing to the storage.  Otherwise, the `save_to_memory` may block the training too long if the time to write to the storage is long.

dlrover/python/elastic_agent/torch/ckpt_saver.py
workingloong: It is difficult to re-initialize a new Saver after the Saver starts the thread to wait the event to save.

workingloong: We cannot re-create a new Saver because the new Saver will clean the tensor meta in the shared dict and the shared memory  with or without the change of `global_shard_num`. If the `global_shard_num` is not changed, the training need to restore the checkpointing state dict from the shared memory with tensor meta.

liyzcj (approved):
lgtm

PR #882, opened 2023-12-11
Start a thread to save the checkpoint before restarting the training.

### LINKED ISSUE #878: Notify other works before rendezvous when training process failed.
For now, when a training process failed, the worker agent will running into rendezvous and change the worker group membership, other workers detect the membership changed and restart training processes.

This logic works well, but when running with async checkpoint and other logic after training process die and before rendezvous, we should notify other workers to do a synchronization.

workingloong:

Start a thread to save the checkpoint before restarting the training.

Fix #878

Avoid the agent blocks too long time when saving the checkpoint.

No.

UT.


liyzcj (approved):
lgtm

PR #884, opened 2023-12-12
Add worker running stage

samplise:

Split running stage to worker_running and ps_running.

The worker init and worker running requires to use different optimizers.

No.

Local test.


workingloong (approved):
LGTM

PR #894, opened 2023-12-17
Fix: fix max-iter setting in nanoGPT example.

Antlera:

Fix max-iter setting in nanoGPT example.

The misplaced breaking condition caused the latest checkpoint stuck at "checkpoint-max-iter" while continuing training.

No.

Test by running nanoGPT example and checking the checkpoint files.

### DIFF
--- a/examples/pytorch/nanogpt/train.py
+++ b/examples/pytorch/nanogpt/train.py
@@ -41,7 +41,7 @@ from dlrover.trainer.torch.elastic.trainer import ElasticTrainer
 os.environ["CUDA_DEVICE_ORDER"] = "PCI_BUS_ID"
 
 # We should use a shared storage to persist the checkpiont.
-checkpoint_dir = "/nas/nanogpt-ckpt/"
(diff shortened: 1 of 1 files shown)


examples/pytorch/nanogpt/train.py
```diff
@@ -377,12 +377,15 @@ def train():
...
                     local_iter_num += 1
-
+                    if iter_num % args.checkpoint_step == 0:
+                        ckpt_manager.save(epoch, iter_num)
                     # Termination conditions
                     if iter_num > max_iters:
                         break
-                    if iter_num % args.checkpoint_step == 0:
-                        ckpt_manager.save(epoch, iter_num)
+                if iter_num > max_iters:
```
workingloong, line 385: The line are  repeated with line 383.

Antlera: This lines are necessary，as there are three layers loop needing break. Anyway，it is not a elegant way to repeat this lines.

workingloong (approved):
LGTM

PR #897, opened 2023-12-18
Support saving FSDP shard state dict.

### LINKED ISSUE #840: Use SHARD_STATE_DICT to save and load the FSDP state dict.
Now, we use the `StateDictType.FULL_STATE_DICT` to save and load the FSDP state dict. However, each rank need to gather all model and optimizer state dict from others which is inefficient and may result in an OOM error.

workingloong:

Use `torch.distributed.checkpoint` to save and load the shard FSDP state dict.

Fix #840 

No.

We can test it by `dlrover-run example/pytorch/nanogpt/train.py`


workingloong:
A new PR in #899

PR #929, opened 2024-01-04
Simplify mnist example.

youxingling:

Simplify mnist example.

Help users run more easily.

No.

UT.


scripts/build_wheel.sh
```diff
@@ -21,5 +21,5 @@ make -f dlrover/Makefile
 
 # Create dlrover package
 echo "Building the wheel for dlrover."
-rm -rf ./build/lib
```
workingloong, line 24: Why do we need to remove the line?

examples/pytorch/mnist/cnn_train.py
```diff
@@ -98,21 +97,20 @@ def train(args):
     """
     setup()
 
-    train_data = torchvision.datasets.ImageFolder(
-        root=args.training_data,
-        transform=transforms.ToTensor(),
+    train_dataset = datasets.MNIST(
```
workingloong, line 100: It is better to use a flag to download the MNIST dateset. Sometimes, it is very slow to download the MNIST dataset. For example:

```Python
if args.training_data:
    train_dataset = torchvision.datasets.ImageFolder(
          root=args.training_data,
          transform=transforms.ToTensor(),
else:
    train_dataset = datasets.MNIST(
        "./data", train=True, transform=transforms.ToTensor(), download=True
    )
```

examples/pytorch/mnist/README.md
```diff
@@ -3,20 +3,6 @@
 The document describes how to use DLRover to train a Pytorch CNN model
 with MNIST dataset.
 
-## Prepare Data
```
workingloong, line 6: It is better to retain the introduction and add a way to create the dataset by `datasets.MNIST()`.

scripts/build_wheel.sh
youxingling: Just to reduce misunderstandings. Before this line, we have already executed `rm -rf build`. And the build file doesn't exist until we execute `python setup.py`.

workingloong: The command will not raise an error even if the `./build/lib` does not exist. So, it is better to retain it to clear the cache.

examples/pytorch/mnist/README.md
```diff
@@ -27,6 +29,13 @@ pip install dlrover -U
 
 Then, we can use `dlrover-run` to start the training by
 
+```bash
+dlrover-run --standalone --nproc_per_node=${GPU_NUM} \
```
workingloong, line 33: Now, we don't need to set `--standalone` to run locally. 
`dlrover-run --nproc_per_node=${GPU_NUM}` is OK.

workingloong (approved):
LGTM

PR #930, opened 2024-01-05
Add a simple example of Flash Checkpoint.

workingloong:


setup.py
```diff
@@ -45,7 +45,7 @@
     url="https://github.com/intelligent-machine-learning/dlrover",
     install_requires=install_requires,
     extras_require=extra_require,
-    python_requires=">=3.8",
+    python_requires=">=3.6",
```
youxingling, line 48: As far as I know, `multiprocessing.shared_memory` is the new in Python3.8. When I run codes in Python3.7, it will report Error: cannot import name ‘shared_memory‘.
Reference: https://docs.python.org/3.8/library/multiprocessing.shared_memory.html?highlight=multiprocessing%20shared_memory#multiprocessing.shared_memory.SharedMemory.close

youxingling: @workingloong

workingloong: Thanks for the comments. I did not notice the problem and I have submitted a PR  #934

PR #933, opened 2024-01-09
Set the node number into the local job manager.

workingloong:

Set the node number into the local job manager.

The local job manager need to collect the error message from nodes.

No.

UT.


Antlera (approved):
LGTM

PR #936, opened 2024-01-09
Fixed issues due to interface differences between tf1.x and tf2.x

mars1248:
Fixed issues due to interface differences between tf1.x and tf2.x


workingloong (approved):
LGTM

PR #939, opened 2024-01-10
WIP: NPU failure node check

samplise:

Run network check before training on npu.

Only run 1 round to check network to avoid restart training errors

No.

Unit test.


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -110,6 +110,7 @@ class ElasticLaunchConfig(LaunchConfig):
     """
 
     network_check: bool = False
+    quick_check: bool = False
```
workingloong, line 113: `check_round_num: int = 2` maybe better. We can set `check_round_num=1` if we want to quickly check the node health.

merlintang (commented):
what is the difference between check NPU vs N-GPU

PR #943, opened 2024-01-10
Develop CheckpointStorage to support different remote storages.

workingloong:

Develop CheckpointStorage to support different remote storages.

Some storage like DFS/HDFS does not support posix interfaces.

No.

UT.


dlrover/python/common/storage.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class PosixDiskStorage(CheckpointStorage):
+    def __init__(self):
+        self._latest_path = ""
+
+    def write(self, content, path):
+        dir = os.path.dirname(path)
+        os.makedirs(dir, exist_ok=True)
+        mode = "w"
```
samplise, line 99: It is better to use constants for mode.

workingloong: The Python native modes are also strings referred to https://docs.python.org/3/library/functions.html#open. It

samplise (approved):

PR #973, opened 2024-01-24
Update the pod attributes if they are not None.

workingloong:

Update the pod attributes if they are not None.

Avoid overwriting the pod metadata.

No.

UT.


samplise (approved):
LGTM

PR #974, opened 2024-01-24
The method to get the resume path from the tracker file.

workingloong:

Return None if the tracker file does not have a step value.

An error will happen if the checkpoint directory is invalid.

No.

UT.


samplise (approved):
LGTM

PR #984, opened 2024-01-31
clear used state dict

FKyms:

clear used state_dict

When ckpt saved, the related state_dict can be cleared, or it will take up a lot of memory.
more than this, if not cleared, the duplicated mem will be copied into child process when evaluate

no


workingloong (approved):
LGTM

PR #991, opened 2024-02-01
An argument to enable saving checkpoint at the breakpoint.

workingloong:

An argument to enable saving checkpoint at the breakpoint.

After a failure occurs, the agent restarts the worker until it finishes saving the checkpoint from the memory to the storage. However, it may takes  a long time to save the checkpoint to storage.

Yes.

UT.


liyzcj (approved):
LGTM

PR #998, opened 2024-02-08
fix saver factory for creation after training process restarts.

liyzcj:

We propose to modify the agent's implementation to use a factory thread that continuously attempts to create the saver, rather than relying on a single creation attempt.

In the current implementation, if the training process fails before the notification for asynchronous saver creation is sent, the saver will not be instantiated on the agent's side. This absence of the saver can result in an error during the next training process. The proposed change aims to ensure that the saver is reliably created, thereby preventing such errors.

No

UT

### DIFF
--- a/dlrover/python/elastic_agent/torch/ckpt_saver.py
+++ b/dlrover/python/elastic_agent/torch/ckpt_saver.py
@@ -411,11 +411,17 @@ class AsyncCheckpointSaver(metaclass=ABCMeta):
         """
 
         def _saver(class_meta: ClassMeta):
+
+            # if the thread is not alive, the saver may be created
+            if cls._saver_instance is not None:
+                cls._saver_instance.close()
+                cls._saver_instance = None
+
             module = importlib.import_module(class_meta.module_path)
             class_def = getattr(module, class_meta.class_name)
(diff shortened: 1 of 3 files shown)


dlrover/python/elastic_agent/torch/ckpt_saver.py
```diff
@@ -409,19 +409,42 @@ def start_async_saving_ckpt(cls):
...
+        def _factory():
+            logger.info("Start the checkpoint saver factory.")
+
+            while True:
+                class_meta: ClassMeta = sq.get()
+
+                # use a lock to avoid concurrent creation of the saver
+                with threading.Lock():
+
+                    if cls._saver_instance is not None:
```
workingloong, line 432: If the thread to save checkpoint fails, the agent can create a new thread. For example:
```Python
saver_thread  = None
if cls._saver_instance is not None and saver_thread and saver_thread.is_alive():
     continue
saver_thread = threading.Thread(
                        target=_saver,
                        args=(class_meta,),
                        name="checkpoint-saver",
                        daemon=True,
                    )
saver_thread.start()
```

workingloong (approved):

PR #1007, opened 2024-02-26
Relaunch th node if the node breaks down when checking the network.

workingloong:

Relaunch the node if the node breaks down when checking the network.

The k8s status may be inconsistent with the node status if the machine crashes.

No.

UT.


samplise (approved):
lgtm

PR #1013, opened 2024-03-01
The job master diagnoses  whether the worker is dead by heartbeat.

### LINKED ISSUE #1010: The Job master monitors the worker status by heartbeat.
# Background

Now, the job master of DLRover monitors the worker status by k8s Pod events. However, the event is lost and the Pod status does not change if the machine of Pod crashes. In this case, The master will not restart the worker because it does not know the worker is dead.  To solve the issue, the worker can periodically send a heartbeat to the job master.  The job master can decide the worker is dead if it does not receive a heartbeat from the worker for a long time.

# Design

- The `ElasticAgent` on a worker Pod starts a thread to periodically send a heartbeat to the job master. The heartbeat message should be extendable because the worker can send other node metrics to the job master in the future.
- The job master decides the worker is dead if the job master has not received a heartbeat from a worker.
- The job master relaunches the worker Pod if the worker is dead.

workingloong:

The job master diagnoses whether the worker is dead by heartbeat.

Fix #1010 

No.

UT.

### DIFF
--- a/dlrover/python/elastic_agent/monitor/training.py
+++ b/dlrover/python/elastic_agent/monitor/training.py
@@ -89,7 +89,7 @@ class TorchTrainingMonitor(Singleton):
         self._resource_monitor.start()
         self._resource_monitor.start_monitor_cpu()
         thread = threading.Thread(
-            target=self._periodically_report_step,
+            target=self._periodically_report,
             name="report-step",
             daemon=True,
         )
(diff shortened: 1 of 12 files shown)


dlrover/python/elastic_agent/monitor/training.py
```diff
@@ -119,9 +119,16 @@ def report_resource_with_step(self):
         except Exception as e:
             logger.warning(e)
 
-    def _periodically_report_step(self):
-        if self._group_rank != 0:
-            return
+    def send_heartbeat(self):
+        try:
+            ts = int(time.time())
```
Antlera, line 124: why should we use a int timestamp here? And `now = time.time()` is not converted to a int.

Antlera: Thanks for explaining, LGTM!

Antlera (approved):
LGTM

PR #1015, opened 2024-03-05
Set an ENV to start the monitor in the agent.

workingloong:

Set an ENV to start the monitor in the agent.

We only need to start the monitor in the node of ElasticJob.

No.

UT.


samplise (approved):
lgtm

PR #1020, opened 2024-03-06
Polish the blog to introduce the fault-tolerance of DLRover.

workingloong:


docs/figures/ft_llm_training/dlrover_failure_recovery.png
samplise, line 1: Some suggestions on this figure:
1. Why use the red arrow for "Detect Node"? Detect node is just an action and may not detect failure nodes.
2. It'd be better to emphasize a new node has replaced the failure node. Maybe a different color?

docs/figures/ft_llm_training/dlrover_node_check.png
samplise, line 1: It'd be better to use the timeline here directly.

nash635 (approved):

Issue #1027, opened 2024-03-11
ATorch pre-commit fails due to mypy


skydoorkai:
Issue:
/github/home/.cache/pre-commit/reponl4tex6o/py_env-python3.10/lib/python3.10/site-packages/requests-stubs/structures.pyi:21: error: Positional-only parameters are only supported in Python 3.8 and greater  [syntax]

Solution:
- update mypy version from 0.900 to 0.981 in atorch/.pre-commit-config.yaml
- 0.981 would check  Unsupported operand types, result in torch_version call result comparison error such as:

atorch/atorch/modules/distributed_modules/mappings.py:17: error: Unsupported operand types for <= ("Tuple[Tuple[int, ...], str]" and "Tuple[int, int, int]")  [operator]

add # type: ignore in all torch_version() comparison commands.

PR #1031, opened 2024-03-13
Update atorch precommit mypy version to v0.981

adamantboy:
See issue #1027, we update atorch mypy version to v0.981 to solve the problem of "Positional-only parameters are only supported in Python 3.8 and greater" during pre-commit.


workingloong:
Please submit a PR from your forked repo. You can see the detail https://github.com/intelligent-machine-learning/dlrover/blob/master/docs/developer_guide.md

PR #1033, opened 2024-03-14
Update atorch precommit mypy version to v0.981

adamantboy:
See issue https://github.com/intelligent-machine-learning/dlrover/issues/1027, we update atorch mypy version to v0.981 to solve the problem of "Positional-only parameters are only supported in Python 3.8 and greater" during pre-commit.


skydoorkai (approved):
lgtm

PR #1064, opened 2024-04-02
Support passing load_strategy to AtorchTrainer.

rockychan724:

Add check for "load_strategy" parameter when user set it.

Support users  passing "load_strategy" parameter to AtorchTainer.

No.

atorch/atorch/tests/common_tests/trainer_test.py


skydoorkai (approved):
lgtm

PR #1074, opened 2024-04-12
Fault Diagnosis Design

samplise:

Add fault diagnosis design documents.

Explain the design of fault diagnosis system.

No

No


liyzcj:
Can we add a new warning level, which will produce warning event for Self-healing system or manual assisted processing？

docs/design/faults-diagnosis.md
```diff
@@ -0,0 +1,83 @@
...
+
+Upon detecting a failure, the diagnosis system must pinpoint its root causes and 
+execute appropriate actions to restore the training process. For example, a brief 
+network jitter causing a network timeout may erroneously flag a machine as faulty, 
+leading to unnecessary replacement. A simple retry could suffice, avoiding the 
+heavier recovery process. To minimize recovery overhead, we propose four levels of 
+diagnosis actions, ranging from light to heavy intervention:
+1. retry, when the error is temporary
+2. restart training process
+3. re-configure training job, e.g., replace a failure machine
```
workingloong, line 34: Does `re-configure training job` means the training job launches a new node to replace a failure node?

docs/design/faults-diagnosis.md
```diff
@@ -0,0 +1,83 @@
...
+4. fail the job and report, when there is a fatal error
```
workingloong, line 35: A full stop is missing at the end of the 4 sentences above.

workingloong (commented):
We need to add an example to explain how to diagnose the hang based on cuda events. In fact, the agent of each node will send the python stack on each rank to the job master if the cuda event of the kernel in each GPU is timeout. The job master does not collect the cuda event but the python stack.

docs/design/faults-diagnosis.md
samplise: Actually, here I indicate any configuration change (e.g., resource or parameter configuration). And the failure nodes replacement is included.

samplise:
> Can we add a new warning level, which will produce warning event for Self-healing system or manual assisted processing？

The analyst can generate the warning events to help the diagnostician diagnose the faults. Detailed interface design will be finalized later.

PR #1092, opened 2024-04-22
The agent waits for async saving checkpoint before exiting.

workingloong:

The agent waits for asynchronously saving checkpoint before exiting.

The thread may be killed even if it is saving the checkpoint from the memory to the storage when the agent exits.

No.

UT.

### DIFF
--- a/dlrover/python/elastic_agent/torch/training.py
+++ b/dlrover/python/elastic_agent/torch/training.py
@@ -83,6 +83,9 @@ except (ModuleNotFoundError, ImportError):  # noqa: F841
 __all__ = ["launch_agent"]
 
 
+_DEFAULT_INTERVAL = 5
+
+
 def _set_paral_config():
     """
     Set up the directory and path for the parallelism configuration.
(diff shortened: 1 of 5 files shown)


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -607,6 +608,15 @@ def _invoke_run(self, role: str = DEFAULT_ROLE) -> RunResult:
             else:
                 raise Exception(f"[{role}] Worker group in {state.name} state")
 
+    def _wait_async_saver(self):
+        saver = AsyncCheckpointSaver.get_ckpt_saver()
+        if saver:
+            while saver.wait_saving_checkpoint():
+                time.sleep(5)
```
BalaBalaYi, line 615: Better use a constant or configurable variable instead of magic number 5.

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -607,6 +608,15 @@ def _invoke_run(self, role: str = DEFAULT_ROLE) -> RunResult:
...
+                logger.info(
+                    "Wait for the asynchronous saver " "saves the checkpoint."
```
BalaBalaYi, line 617: lint here. Or we can add  'the time that has been waited' in the logging?

BalaBalaYi (approved):
LGTM

PR #1103, opened 2024-05-06
Refactor the job collector to collect the count of failover.

workingloong:

Refactor the job collector to collect the count of failover.

To observe the benefits of DLRover failover.

No.

UT.

### DIFF
--- a/dlrover/python/master/stats/job_collector.py
+++ b/dlrover/python/master/stats/job_collector.py
@@ -72,6 +72,14 @@ class BaseMetricCollector(metaclass=ABCMeta):
     ):
         pass
 
+    @abstractmethod
+    def collect_custom_data(self, metric_dict):
+        pass
+
+    @abstractmethod
+    def collect_job_exit_reason(self, reason):
+        pass
+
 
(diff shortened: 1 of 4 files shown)


dlrover/python/master/stats/job_collector.py
```diff
@@ -125,7 +133,9 @@ def _report_runtime_stats(self):
         self._stats_reporter.report_runtime_stats(self._runtime_metric)
 
     @BaseMetricCollector.catch_exception
-    def collect_custom_data(self):
+    def collect_custom_data(self, metric_dict=None):
+        if metric_dict and not metric_dict <= self._custom_metric:
```
samplise, line 137: Can we compare two dictionaries like dict1 < dict2? And why `metrics_dict > self._custom_metrics` directly?

workingloong: In Python 3, you can use dict.items() to get a set-like view of the dict items. You can then use the <= operator to test if one view is a "subset" of the other.

samplise (approved):

PR #1111, opened 2024-05-09
Refactor /examples/pytorch/NanoGPT/*_train.py

Filtee:

I refactored example/pytorch/NanoGPT/*_train.py, i.e. train.py, ds_train.py, and fsdp_train.py.

These three files have almost identical code, with the main difference being the technology they use: fsdp, deepspeed, or DDP, which would just lead to a tiny little difference in code.

So I just decided to refactor all of them anyway. It seems that all of them works fine.

Nope. It just relates to the example showing, not the main part of DLRover.

It works fine…

### DIFF
--- a/examples/pytorch/nanogpt/train.py
+++ b/examples/pytorch/nanogpt/train.py
@@ -15,13 +15,13 @@
 The start command on a local ndoe:
 
 dlrover-run --nproc_per_node=2 train.py \
-    --n_layer 48 --n_head 16 --n_embd 1600 --data_dir './' \
+    --n_layer 48 --n_head 16 --n_embd 1600 --data_dir "./" \
     --epochs 50 --save_memory_interval 50 --save_storage_interval 500
 """
 
-
 import argparse
 import contextlib
+import functools
 import os
 import time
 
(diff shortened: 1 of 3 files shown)


examples/pytorch/nanogpt/train.py
```diff
@@ -328,16 +309,59 @@ def wrapper(*args, **kwargs):
...
+    """
+    model = model_params["model"]
+    optimizer = model_params["optimizer"]
+    train_loader = model_params["train_loader"]
+    checkpointer = ckpt_params["checkpointer"]
+    checkpoint_dir = ckpt_params["checkpoint_dir"]
+    os.makedirs(checkpoint_dir, exist_ok=True)
+
+    if ckpt_params["use_native"]:
+        ckpt_dict = torch.load(os.path.join(checkpoint_dir, "50.pt"))
```
workingloong, line 346: We should select a checkpoint file from the checkpoint directory not use a fixed`50.pt`.

Filtee: Yep, I also find this line of code a bit weird, but it's how the original source code was before my refactoring. 😯

(The following code comes from the original code.)

截屏2024-05-10 14 56 53

Should I make some modifications to it?

Filtee: Alright, I fixed it anyway.

workingloong (approved):
LGTM

PR #1125, opened 2024-05-15
fix pod event converting invocation

BalaBalaYi:

Bug fix.

Should use 'k8s client' as parameter.

Nope.

UT.


workingloong (approved):
LGTM

PR #1141, opened 2024-05-27
Fix the example of deeprec to train the deepfm model with CRITEO dataset.

workingloong:

Fix the example of deeprec to train the deepfm model with CRITEO dataset.

The example to use DLRover to train TensorFlow Models.

No.

`kubectl -n dlrover apply -f  examples/tensorflow/criteo_deeprec/manual_job.yaml`


hxdtest (approved):
LGTM

PR #1163, opened 2024-06-19
optimize svs operation

BalaBalaYi:

1. set default port name in port spec: would fail if port name is none when doing patch operation
2. support patch skipping: no need to patch svc if selector is the same

To enhance svc operation.

No.

UT.


cailun01 (approved):
LGTM

PR #1166, opened 2024-06-23
Modify the controller's resource allocation settings.

Filtee:

I modified the controller's resource allocation settings, with help from @Antlera 

And... I fixed some TYPO during reading codes.

Otherwise, it will experience instability when deployed in the cluster.

Nope.

Nope.


dlrover/go/operator/config/manifests/bases/deployment.yaml
```diff
@@ -66,10 +66,10 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 128Mi
+            memory: 4090Mi
```
workingloong, line 69: The memory is too large if we deploy the controller in a minikube cluster.

dlrover/python/common/multi_process.py
```diff
@@ -141,7 +141,7 @@ class LockAcquireResponse(SocketResponse):
     A response to acquire a shared lock using local socket.
 
     Attributes:
-        acquired (bool): Ture if the lock is acquired.
+        acquired (bool): True if the lock is acquired.
```
linkerzhang, line 144: please move all these kind of "wording" changes / fixes into a separate PR.

PR #1173, opened 2024-07-01
Pod scaler enhancement: support concurrent creation

BalaBalaYi:

1. Use 'deque' to replace 'list' for better performance and thread safe.
2. Support pod creation in concurrency.
3. Optimize function's input fields for some input can be used from class object directly.
4. Separate '_create_pod_from_queue' function for better adaption with internal repo's implementation.

To accelerate pod creation.

Nope.

UT.


BalaBalaYi:
This PR need internal compatible work. Considering the internal release, therefore the merge of the current PR must be delayed.

workingloong:
We can use volcano job to create Pods. Volcano use goroutines to speed up creating Pods https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/podgroup/pg_controller.go#L135-L137.

BalaBalaYi:
> We can use volcano job to create Pods. Volcano uses goroutines to speed up creating Pods https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/podgroup/pg_controller.go#L135-L137.

Will do. These two things are actually no conflict.

samplise (approved):

PR #1181, opened 2024-07-08
Improve training port conflict avoid

samplise:

We can configure the start port of the training job.

1. It is more efficient to configure the start training port when the users know the available port range.
2. We can disable the port conflict-avoiding function.

No.

Unit test.

### DIFF
--- a/dlrover/python/common/grpc.py
+++ b/dlrover/python/common/grpc.py
@@ -107,6 +107,8 @@ def find_free_port_in_set(ports):
 def find_free_port_for_hccl(start=60000) -> int:
     cur_start = start
     end = start + 10000
+    if end > 65000:
+        end = 65000
     logger.info(f"Try to find available port for hccl from {start}")
     while True:
         try:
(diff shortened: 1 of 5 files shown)


majieyue (approved):
use training_port as a switch, LGTM~

dlrover/python/common/grpc.py
```diff
@@ -104,8 +104,9 @@ def find_free_port_in_set(ports):
     raise RuntimeError(f"Fail to find a free port in {ports}")
 
 
-def find_free_port_for_hccl(start=60000, end=70000) -> int:
+def find_free_port_for_hccl(start=60000) -> int:
     cur_start = start
+    end = start + 10000
```
BalaBalaYi, line 109: The max port number is 65535. So ```end = start + 5000```

BalaBalaYi (changes requested):

BalaBalaYi (approved):
LGTM

PR #1185, opened 2024-07-09
Skip restart training process on failure nodes

samplise:

When observing a failure node from training logs, the dlrover skips training process restart and relaunch new worker directly.

It fails to resume training via restarting training process on a failure node.

No.

Unit test.


dlrover/python/common/diagnosis.py
```diff
@@ -50,6 +52,7 @@ def get_type(self) -> str:
 
 class TrainingLog(DiagnosisData):
     def __init__(self, timestamp: int):
+        super().__init__()
         if timestamp == 0:
```
BalaBalaYi, line 56: Should define following in the abstract class 'DiagnosisData':
```
if timestamp == 0:
    self.timestamp = int(round(datetime.now().timestamp()))
else:
    self.timestamp = timestamp
```

dlrover/python/common/diagnosis.py
```diff
@@ -74,3 +77,17 @@ def get_timestamp(self) -> int:
...
     def get_type(self) -> str:
         return DiagnosisDataType.CHIPMETRICES
+
+
+def should_relaunch_worker(log_file: str) -> bool:
+    if len(log_file) == 0:
+        return False
+    try:
+        with open(log_file, "r") as f:
+            error = "error code is 507035"
```
BalaBalaYi, line 87: These kind of error should be defined as constant in a independent class or module. There will certainly be more such error info in the future.

dlrover/python/elastic_agent/datacollector/log_collector.py
```diff
@@ -19,6 +19,7 @@
 
 class LogCollector(DataCollector):
     def __init__(self, *args, **kwargs):
+        super().__init__()
         pass
```
BalaBalaYi, line 23: remove the 'pass'

dlrover/python/tests/test_diagnosis_common.py
```diff
@@ -0,0 +1,32 @@
...
+# limitations under the License.
+
+import os
+import unittest
+
+from dlrover.python.common.diagnosis import should_relaunch_worker
+
+
+class DiagnosisTest(unittest.TestCase):
+    def setUp(self):
```
BalaBalaYi, line 21: Just remove the unused function.

dlrover/python/tests/data/training.log
```diff
@@ -0,0 +1 @@
+error code is 507035
```
BalaBalaYi, line 1: Better intercept part of the original log as test data and simulate the actual scenario as much as possible.

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -629,7 +632,10 @@ def _invoke_run(self, role: str = DEFAULT_ROLE) -> RunResult:
                 logger.error(f"The worker fails with {run_result.failures}")
                 self._report_failure_to_master(run_result.failures)
                 self._save_ckpt_to_storage()
-                if self._remaining_failovers > 0:
+                if (
+                    self._remaining_failovers > 0
```
BalaBalaYi, line 636: Should put the 'self._remaining_failovers > 0' logic into 'shoud_relaunch_worker' function.

BalaBalaYi (changes requested):

BalaBalaYi (approved):

PR #1200, opened 2024-07-17
Fix heartbeat when there is node relaunched.

BalaBalaYi:

1. When new node is relaunched, the new node object should reset the heartbeat to 0.
2. Optimize the logging.

Fix heartbeat issue when there is node relaunched.

No.

UT and full training test.

### DIFF
--- a/dlrover/python/common/node.py
+++ b/dlrover/python/common/node.py
@@ -270,6 +270,7 @@ class Node(object):
         new_node.start_time = None
         new_node.create_time = None
         new_node.finish_time = None
+        new_node.heartbeat_time = 0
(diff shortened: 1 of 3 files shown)


dlrover/python/common/node.py
```diff
@@ -270,6 +270,7 @@ def get_relaunch_node_info(self, new_id):
         new_node.start_time = None
         new_node.create_time = None
         new_node.finish_time = None
+        new_node.heartbeat_time = 0
```
BalaBalaYi, line 273: Key part.

majieyue (approved):
lgtm

PR #1207, opened 2024-07-23
Optimize node status from pod phase.

BalaBalaYi:

Set to 'Deleted' status if the pod has 'deletion_timstamp' when listing pod from k8s.

To prevent repeated worker heartbeat warning when pod deletion failed.

A example scene:
1. the heartbeat is working
2. a worker heartbeat lost and timeout
3. master relaunch the worker who lost heartbeat
4. master receive to_deleted event and update node status to deleted, but k8s failed to delete the pod(failure node)
5. master can still get pod info from k8s and the pod.status.phase is running but with deletion_timestamp
6. for the node status is based on the pod.status.phase, the node status rewrite to running again
7. master will always find that worker who lost heartbeat (never end until the pod is deleted manually)

No.

UT.


workingloong (approved):
LGTM

PR #1209, opened 2024-07-24
Keep conflict processing when env set

BalaBalaYi:

1. Use default env's port as start port.
2. Update max port to 65500.
3. Optimize ut for higher efficiency.

Keep conflict processing when env has already been set.

No.

UT.


majieyue (approved):
lgtm

PR #1212, opened 2024-07-26
Support action timeout processing.

BalaBalaYi:

1. Add new logic in torch-training-monitor to judge 'is action timeout'.
2. Add time setting around 'stop' action to record cost.
3. Will raise error(pod failover) if there is action timeout.

No.

UT.


majieyue (approved):
LGTM

we can optimize it by a more flexible stop_workers implementation

PR #1217, opened 2024-07-30
Fix scaler async execution.

BalaBalaYi:

Do sync operation(pop element from deque) before async execution.

To resolve invalid async execution calls.

No.

UT.


majieyue (approved):
lgtm

PR #1220, opened 2024-08-01
Optimize pending judgement: when all nodes pending

BalaBalaYi:

Pass the current total worker number to the function to cover the case: 'all nodes pending'.

Need cover the following case when doing pending judgement:
All nodes pending at beginning(min nodes required is None).

No.

UT and full training test.


majieyue (approved):
LGTM

PR #1229, opened 2024-08-08
Fix flash ckpt

BalaBalaYi:

1. Annotations fixed.
2. Wrong variable fixed.
3. Add timeout optional for fsdp.
4. Use ckpt's timeout as mkdir's timeout.
5. Remove the tmp dir at the beginning for the history content may cause problems.

Multi bug fix and enhancement.

No.

UT.


majieyue (approved):
lgtm

PR #1238, opened 2024-08-15
Skip 'should early stop' for non all reduce job.

BalaBalaYi:

To skip 'should early stop' for non all reduce job.

Optimization for different case.

No.

UT.

### DIFF
--- a/dlrover/python/master/node/dist_job_manager.py
+++ b/dlrover/python/master/node/dist_job_manager.py
@@ -217,13 +217,10 @@ class DistributedJobManager(JobManager):
         return self._job_resource.worker_num
 
     def is_all_reduce_type_job(self):
-        if (
+        return (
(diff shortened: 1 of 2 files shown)


dlrover/python/master/node/dist_job_manager.py
```diff
@@ -216,6 +216,15 @@ def _has_running_workers(self):
     def get_worker_num(self):
         return self._job_resource.worker_num
 
+    def is_all_reduce_type_job(self):
+        if (
```
samplise, line 220: `return self._job_args.distribution_strategy == DistributionStrategy.ALLREDUCE`

BalaBalaYi: done

samplise (approved):
lgtm

PR #1250, opened 2024-08-23
Sse rdzv timeout as insufficient timeout

BalaBalaYi:

1. Worker report 'join timeout' by ```report_rdzv_params```.
2. Master saves the 'join timeout' in worker manager.
3. Master base on the 'join timeout' to setup insufficient timeout.

For back compatibility(worker who don't report the join timeout):
Use the default 600s as the 'join timeout'.

For protection:
The timeout range is (600s, 3600s).

Previously, we used the pending timeout period as the criterion for determining insufficiency. However, considering that users may configure the networking timeout period themselves, it is more scientific to directly use the networking timeout period for this configuration. 

The current definition of this timeout period is **1.5 times the networking timeout period**, which is used to avoid scenarios where **errors cannot be recovered within a complete round of the networking timeout period**.

No.

UT.

### DIFF
--- a/dlrover/python/common/constants.py
+++ b/dlrover/python/common/constants.py
@@ -310,6 +310,10 @@ class CheckpointConstant(object):
     SAVE_TIMEOUT = 600
 
 
+class RdzvConstant(object):
+    JOIN_TIMEOUT = 600
+
+
 class Accelerators(object):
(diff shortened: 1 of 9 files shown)


majieyue (approved):
lgtm

PR #1254, opened 2024-08-29
Fix serveral  issue when using fsdp checkpointer.

BalaBalaYi:

1. Delay ckpt_saver config updating to avoid socket client creation before socket server creation.
2. Other optimization.

Fix unexpected error during checkpointing.

No.

UT and training test.


majieyue (approved):
lgtm

PR #1268, opened 2024-09-12
Make diagnosis agent singleton

samplise:

Make diagnosis agent to be singleton.

Avoid to create duplicate diagnosis agent on the same worker.

No.

Unit test.


workingloong (approved):
LGTM

PR #1269, opened 2024-09-12
Filter non training logs

samplise:

When to read the training logs, DLrover only reads logs after the training is started.

Avoid non-training log noises.

No.

Unit test.


dlrover/python/diagnosis/datacollector/training_log_collector.py
```diff
@@ -32,6 +32,15 @@ def collect_data(self) -> TrainingLog:
             return TrainingLog()
         byte_logs = read_last_n_lines(self._log_file, self._n_line)
         logs = [str(line) for line in byte_logs]
+        # only collect training logs
+        if len(logs) > 0:
```
BalaBalaYi, line 36: add a ut

BalaBalaYi (approved):
LGTM

PR #1271, opened 2024-09-12
Skip empty error codes when check node failures

samplise:

Check if error code is empty when detecting failure node.

There is a bug when error code is empty.

No.

Unit test.


BalaBalaYi (approved):
LGTM

PR #1277, opened 2024-09-24
Worker get elastic run config from master

samplise:

Before starting training, workers read elastic run configures from the master.

We can distribute run configures from the master.

No.

Unit test.

### DIFF
--- a/dlrover/trainer/torch/elastic_run.py
+++ b/dlrover/trainer/torch/elastic_run.py
@@ -304,32 +304,85 @@ def _elastic_config_from_args(
     args,
 ) -> Tuple[ElasticLaunchConfig, Union[Callable, str], List[str]]:
     config, cmd, cmd_args = config_from_args(args)
+
+    master_config = _elastic_config_from_master(config)
     elastic_config = ElasticLaunchConfig(**config.__dict__)
 
     # PyTorch >= 2.3.0 remove log_dir in the LaunchConfig.
     if not version_less_than_230():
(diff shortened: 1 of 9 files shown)


workingloong:
Does the PR support updating the training configurations without restarting the job?

samplise:
> Does the PR support updating the training configurations without restarting the job?

No. This pr allows to configure all jobs in a cluster with the same configuration in a simple way. Otherwise, we have to ask all users to have the same configuration in their dlrover-run command line.

majieyue (approved):
lgtm

dlrover/trainer/torch/elastic_run.py
```diff
@@ -304,26 +304,46 @@ def _elastic_config_from_args(
...
+
+    master_config = _elastic_config_from_master(config)
     elastic_config = ElasticLaunchConfig(**config.__dict__)
 
     # PyTorch >= 2.3.0 remove log_dir in the LaunchConfig.
     if not version_less_than_230():
         elastic_config.log_dir = config.logs_specs.root_log_dir
+
     elastic_config.network_check = getattr(args, "network_check", False)
+    if master_config.network_check:
```
BalaBalaYi, line 316: No need to set 'false' if config from master is false?

samplise: Yes. Here we assume the local command has the highest priority.

BalaBalaYi (approved):
LGTM

PR #1293, opened 2024-10-12
add exception handler in _get_master_addr_port since the port might b…

majieyue:

add try-exception to protect _get_master_addr_port and add retry of the calling to relieve the port sync problem

the return value of _get_master_addr_port may be invalid

No

UT

### DIFF
--- a/dlrover/python/elastic_agent/torch/training.py
+++ b/dlrover/python/elastic_agent/torch/training.py
@@ -485,7 +485,7 @@ class ElasticTrainingAgent(LocalElasticAgent):
                     spec.master_port,
                 )
 
-        master_addr, master_port = self._get_master_addr_port(store)
+        master_addr, master_port = self._safe_get_master_addr_port(store)
 
         # compatible with torch 2.4
         if not version_less_than_240():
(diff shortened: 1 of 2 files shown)


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -543,6 +543,19 @@ def _get_master_addr_port(self, store: Store) -> Tuple[str, int]:
         master_port = int(store.get("MASTER_PORT").decode(encoding="UTF-8"))
         return (master_addr, master_port)
 
+    def _safe_get_master_addr_port(self, store: Store) -> Tuple[str, int]:
+        for i in range(1, 5):
```
workingloong, line 547: `for _ in range(5)` is ok.

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -543,6 +543,19 @@ def _get_master_addr_port(self, store: Store) -> Tuple[str, int]:
...
+            try:
+                addr, port = self._get_master_addr_port(store)
```
workingloong, line 549: `return self._get_master_addr_port(store)`

workingloong (approved):
LGTM

PR #1295, opened 2024-10-14
Catch exception when get config from brain

samplise:

Add exception catch in get config.

Avoid errors when fail to get configs from master and brain.

No.

Unit test.


dlrover/python/master/node/training_node.py
```diff
@@ -470,6 +470,8 @@ def sync_node_training_port(self, node_id, port) -> SyncNodeTrainingPorts:
                 )
 
     def get_elastic_run_configs(self) -> Dict[str, str]:
+        logger.info("try to collect elastic run configs externally")
```
BalaBalaYi, line 473: Useless logging.

BalaBalaYi (approved):
LGTM

PR #1296, opened 2024-10-16
Fix relaunch node's relaunch limit.

BalaBalaYi:

Relaunch node will inherit the relaunch limit value.

Worker number less 3 should be able to relaunch 3 times.

No.

UT and training.


samplise (approved):
lgtm

PR #1301, opened 2024-10-18
Fix/Enhance node management.

BalaBalaYi:

1.  Add 'deletion_timestamp' condition when judging whether there is another same pod running. To avoid mistakenly identifying the current pod(to be removed) as the target.
2. Still need judgement when there is 0 pod list from k8s.

Fix/Enhance node management.

No.

UT and training.


samplise (approved):
lgtm

PR #1312, opened 2024-10-28
Fix: Use return instead of break to correctly exit loop on path exist…

jinqinn:

1. Add `generateEmbeddedObjectMeta` option to `controller-gen` to support generating metadata for embedded objects.
2. Fix incorrect import of the `megatron` package.
3. Replace break with return when checking for the existence of step_done_dir in storage.

1. This allows CRD developers to include standard Kubernetes metadata (like name, labels, and annotations) in nested fields, providing better support for complex resource structures .
2. Correcting this import ensures that the application can locate and utilize the package correctly, preventing potential runtime errors and improving code stability.
3. Using return allows the function to exit immediately once done_dir is found, avoiding unnecessary iterations and enhancing code readability and efficiency.

NO

YES


dlrover/python/elastic_agent/torch/ckpt_saver.py
```diff
@@ -923,6 +923,7 @@ def save_step_checkpoint(self, step: int):
                 f"Rank {self._node_rank} save checkpoint failed for "
                 f"step {step}"
             )
+            self._writing_storage = False
```
BalaBalaYi, line 926: This modification should be sync to the ```TempDirCheckpointSaver```.

samplise (approved):
lgtm

PR #1317, opened 2024-11-04
Fix empty node issue after master failover

BalaBalaYi:

Re add relaunch worker node info back to ```job_nodes``` in ```JobManager``` during 'list' operation.

Fix empty node issue after master failover.

No.

UT and training(with worker failover + master failover).


samplise (approved):
lgtm

PR #1320, opened 2024-11-05
fix: typo RayJobSubmitter in ray_job_submitter.py

Peefy:

I discovered this error while browsing and learning the code, so I raised a PR to fix it

Typos in code can affect people's learning and use of it

Yes, this means that once people have used it in past versions, they also need to modify it together.

Unit test in CI.


workingloong (approved):
LGTM

PR #1328, opened 2024-11-11
Function using/naming optimized based on job context.

BalaBalaYi:

Function renaming in ```TranningNodeManager```.

To distinguish the usage of different types of methods.

No.

UT.


majieyue (approved):
lgtm

PR #1330, opened 2024-11-12
Delete pod when list pod already succeeded.

BalaBalaYi:

Delete pod if status is 'Succeeded' in positive pod list operation.

Considering that the majority of opinions in the current Kubernetes community are that the resource usage of completed pods should be taken into account, and since the processes and containers have already exited, 'Succeeded' pods do not need to be retained. Therefore 'Succeeded' pods should be deleted.

No.

UT.


workingloong (approved):
LGTM

PR #1339, opened 2024-11-15
Fix the issue that len(indices) and num_samples might not be equal 

sunjq1:

 Modified the definition of `total_size` in the `load_state_dict` function and the definition of `indices` in the `__iter__` function to ensure that `assert len(indices) == self.num_samples`.

 In the previous ElasticDistributedSampler code, the issue where `completed_num` not being divisible by `num_replicas` could cause an `AssertionError` in the scenario of importing a checkpoint was not considered.

No.

UT and training test.


BalaBalaYi (changes requested):
please add ut to cover the issue mentioned

sunjq1:
> please add ut to cover the issue mentioned

 
since I adjusted the logic for splitting indices after loading the checkpoint, I corrected the num values in the UT file

BalaBalaYi (approved):
LGTM

dlrover/trainer/tests/torch/elastic_sampler_test.py
```diff
@@ -58,7 +58,7 @@ def test_checkpoint(self):
         )
         sampler.load_state_dict(sampler_state)
         val = next(iter(sampler))
-        self.assertEqual(val, 64)
+        self.assertEqual(val, 66)
```
zhengchenyu, line 61: This should be caused by scaling. This unit-test tells us that data will be lost.

PR #1340, opened 2024-11-16
fix a bug in infer method

jlsong01:

a bugfix

the parameter of operator.infer() SHOULD be a list with one element [inference], while NOT the whole list of inferences, because inferences may include the inference which does not match the currrent operator.

No

pre-commit test and unittest both passed


dlrover/python/diagnosis/inferencechain/inference_chain.py
```diff
@@ -46,7 +46,7 @@ def infer(self) -> List[Inference]:
                     operator = self.get_operator(inference)
                     if not operator:
                         continue
-                    infs = operator.infer(inferences)
+                    infs = operator.infer([inference])
```
BalaBalaYi, line 49: @samplise have a look

samplise:
This implementation will be more accurate in some cases. For example, we have observed two problems a and b during training. We want to identify the root causes for problem a. However, the final results may differ for only observed a from observed a and b together. Therefore we have all problems as the input for more accurate analysis.

dlrover/python/diagnosis/inferencechain/inference_chain.py
samplise: This implementation will be more accurate in some cases. For example, we have observed two problems a and b during training. We want to identify the root causes for problem a. However, the final results may differ for only observed a from observed a and b together. Therefore we have all problems as the input for more accurate analysis.

workingloong: @samplise how about the comment?

samplise (approved):

PR #1341, opened 2024-11-18
Skip memory limitation for gpu type node relaunch operation.

BalaBalaYi:

1. Skip the memory limitation judgement if node has gpu resource in relaunch operation.
2. Add a default limitation for memory adjustment for 'oom case' under torch training. 

There is a memory limitation to avoid invalid resource scaling(out of quota)  under tensorflow training case. This limitation should not be involved in torch training for now for all the gpu type trainings nowadays are not scalable.

No.

UT.


majieyue (approved):
lgtm

PR #1347, opened 2024-11-18
correct log and complete ut for error_monitor

jlsong01:

code enhancement

1. make log output more clear, since in k8s world you can ONLY mark a worker NODE unschedulable, while not a POD.

1. complete the unit test for error monitor

no

pre-commit test and unittest passed

### DIFF
--- a/dlrover/python/master/monitor/error_monitor.py
+++ b/dlrover/python/master/monitor/error_monitor.py
@@ -151,5 +151,5 @@ class K8sJobErrorMonitor(ErrorMonitor):
         if self.cordon_node_eanbled:
             succeed = self._k8s_client.cordon_node(node.host_name)
             if succeed:
-                logger.info(f"Node {node.name} is marked unschedulable.")
+                logger.info(f"Host {node.host_name} is marked unschedulable.")
         return True
--- a/dlrover/python/master/monitor/error_monitor.py
+++ b/dlrover/python/master/monitor/error_monitor.py
@@ -100,10 +100,10 @@ class SimpleErrorMonitor(ErrorMonitor):
 class K8sJobErrorMonitor(ErrorMonitor):
     """The monitor logs the error data."""
 
-    def __init__(self, namespace="", cordon_node_eanbled=False):
+    def __init__(self, namespace="", cordon_node_enabled=False):
         from dlrover.python.scheduler.kubernetes import k8sClient
 
-        self.cordon_node_eanbled = cordon_node_eanbled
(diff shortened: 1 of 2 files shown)


dlrover/python/master/monitor/error_monitor.py
```diff
@@ -151,5 +151,5 @@ def _handle_node_error(self, node: Node, error_data: str):
         if self.cordon_node_eanbled:
             succeed = self._k8s_client.cordon_node(node.host_name)
             if succeed:
-                logger.info(f"Node {node.name} is marked unschedulable.")
+                logger.info(f"Host {node.host_name} is marked unschedulable.")
```
BalaBalaYi, line 154: 'Node' is a domain model in DLRover. Also, the 'host_name' can be a long string(including idc and zone info), is not readable for user.

dlrover/python/tests/test_error_monitor.py
```diff
@@ -39,9 +39,10 @@ def test_process_error(self):
         )
         self.assertTrue(relaunched)
 
-        err_monitor.process_error(
+        relaunched = err_monitor.process_error(
```
BalaBalaYi, line 42:
```
self.assertFalse(err_monitor.process_error(xxx))
```

BalaBalaYi (dismissed):

dlrover/python/master/monitor/error_monitor.py
jlsong01: Yes, Node is a domain model, however, for K8s platform, the domain model is implemented by K8s Pod. Since the POD is ephemeral, we can not mark a pod as unschedulable, we can only mark the host of a pod as unschedulable. 
So, the origianl log here might be confusing.

dlrover/python/tests/test_error_monitor.py
jlsong01: fixed, thanks

dlrover/python/master/monitor/error_monitor.py
BalaBalaYi: U r right, the target model here is the K8s Node.

dlrover/python/master/monitor/error_monitor.py
```diff
@@ -151,5 +151,5 @@ def _handle_node_error(self, node: Node, error_data: str):
         if self.cordon_node_eanbled:
```
BalaBalaYi, line 151: Could u please correct the spells here?
```eanbled```

workingloong (approved):
LGTM

PR #1352, opened 2024-11-20
add customized step collector into dlrover

majieyue:
usage: add such parameters into agent command line, e.g. dlrover-run

--training_log_file specifies which logfile we collect step info 
--step_rank specifies the rank which will output the step info into logfile 
--step_pattern specifies the regex pattern of step info matching each log line

This patch support user defined step log format, so dlrover can collect training steps info from log file. It is useful for
us to detect training errors, e.g. hang or stepback

step pattern and file match

add customized step collector into dlrover

Yes

UT

### DIFF
--- a/dlrover/python/common/grpc.py
+++ b/dlrover/python/common/grpc.py
@@ -239,6 +239,13 @@ class GlobalStep(Message):
 
 
 @dataclass
+class UserStep(Message):
+    timestamp: int = 0
+    step: int = 0
+    total: int = 1
+
+
+@dataclass
 class HeartBeat(Message):
     timestamp: int = 0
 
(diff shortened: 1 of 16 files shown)


dlrover/python/common/grpc.py
```diff
@@ -238,6 +238,13 @@ class GlobalStep(Message):
     elapsed_time_per_step: float = 0.0
 
 
+@dataclass
+class UserStep(Message):
```
BalaBalaYi, line 242: An advice:
1. Use ```DiagnosisReportData``` instead of the new definition here.
2. Use 'type' to tag it as 'UserStep' info and do resolving in master.

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -177,15 +177,21 @@ class ElasticLaunchConfig(LaunchConfig):
     tee: Union[Std, Dict[int, Std]] = Std.NONE
     training_log_file: str = ""
     failure_node_errors: str = ""
+    step_rank: int = -1
```
BalaBalaYi, line 180: Use a class object here to unify step info. Or u can define a more common class to include these fields.

PR #1357, opened 2024-11-26
Fix unexpected heartbeat issue after worker relaunching.

BalaBalaYi:

1. Node's 'reported status' should not be inherited during relaunching.
2. Revert heartbeat timeout back to 10mins(Considering that some other optimizations have been applied recently, a 15-minute timeout period is somewhat too long.).

Bug fix and enhancement.

No.

UT and Training.

Training validation: failover worker-0 -> manual hang worker-1 -> verify if worker-1 will relaunch by heartbeat timeout


samplise (approved):
lgtm

PR #1361, opened 2024-11-27
Fix socket client using(before socket server creation).

BalaBalaYi:

Create socket client after socket server is ready.

To avoid socket client creation before server creation.

No.

UT.


dlrover/trainer/torch/flash_checkpoint/engine.py
```diff
@@ -133,6 +133,21 @@ def start_saver_process():
     return None
 
 
+def wait_socket_server(socket_server, timeout=60):
```
workingloong, line 136: Add a typing class for socket_server.

BalaBalaYi: done

workingloong (approved):
LGTM

PR #1371, opened 2024-12-03
Do not clean up checkpoint path with step 0

sunjq1:
 
Optimize the handling of boundary values in the checkpoint cleanup path.

During training, step 0 checkpoint are not saved, but the original logic attempts to clean up `/tmp/checkpoint/0`, which triggers the following warning
```
[INFO] [storage.py:258:clean_up] Clean path /tmp/checkpoint/0
[WARNING] [storage.py:261:clean_up] Fail to clean path /tmp/checkpoint/0!
```

No.

UT and training test.


workingloong (approved):
LGTM

PR #1373, opened 2024-12-04
no need to detect cpu hang in torch training,

majieyue:
since cpu may drop to less than 2 percent in the normal training process

only check cpu hang in non-allreduce job

in some allreduce job cases, cpu may drop to less than 2% while the training is normal

No

UT

### DIFF
--- a/dlrover/python/master/node/dist_job_manager.py
+++ b/dlrover/python/master/node/dist_job_manager.py
@@ -1066,14 +1066,15 @@ class DistributedJobManager(JobManager):
             return
         node.update_resource_usage(cpu, memory, gpu_stats)
         cpu_percent = node.used_resource.cpu / node.config_resource.cpu
-        if cpu_percent < _dlrover_context.hang_cpu_usage_percentage:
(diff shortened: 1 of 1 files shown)


dlrover/python/master/node/dist_job_manager.py
```diff
@@ -1066,14 +1066,15 @@ def update_node_resource_usage(
...
         cpu_percent = node.used_resource.cpu / node.config_resource.cpu
-        if cpu_percent < _dlrover_context.hang_cpu_usage_percentage:
-            if node.start_hang_time == 0:
-                now = datetime.now()
-                node.start_hang_time = now.timestamp()
-        else:
-            if node.start_hang_time > 0:
-                now = datetime.now()
-            node.start_hang_time = 0
+        if not self.is_all_reduce_type_job():
```
BalaBalaYi, line 1069: The torch training does not necessarily have to run on the GPU; it can also run on the CPU.

BalaBalaYi:
Torching training != CPU training. So just please keep this impl for now.

PR #1376, opened 2024-12-06
Fix node event process issue.

BalaBalaYi:

Need add 'replica-index' in label selector to identify the unique node to avoid: the logic matches a newly relaunched node .

Bug fix.

No.

UT and training.


majieyue (approved):
lgtm

PR #1377, opened 2024-12-06
Fix the issue of ZeroDivisionError when the CPU request is not configured

sunjq1:
 
This PR enhances the code's robustness by throwing a warning when the CPU request value is not configured

 If the CPU request value is not configured or configure to 0 in the yaml file, the code will trigger a ZeroDivisionError when executing `cpu_percent = node.used_resource.cpu / node.config_resource.cpu`

No

UT and training test.


workingloong (approved):
LGTM

PR #1383, opened 2024-12-09
print exception info in retry_grpc_request

majieyue:

print exception instead of only failure string

debug use

No

UT


samplise (approved):
lgtm

PR #1385, opened 2024-12-11
Fix rendezvous out of sync error

majieyue:

the _check_round is hardcoded and in some corner cases, the rendezvous will never end. this patch will fix it.

If some workers happened to fail in rendezvous procedure, the new worker will re-join the rendezvous from NodeCheckRendezvous phase. Suppose remaining workers are still in the ElasticTrainingRendezvous phase, 
they will raise a RendezvousOutSyncError exception.

In some corner cases, e.g. there are always stragglers in each round and exclude_straggler not set, every workers will run 2 rounds to recheck either fault nodes or straggler nodes. In such cases, supposed new worker join the rendezvous, e.g. the NodeCheckRendezvous of remaining workers, right after the 1st round and before the 2nd round, it will lead to an endless rendezvous, as long as stragglers remains

No

UT and BVT


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -827,6 +830,16 @@ def _initialize_workers(self, worker_group):
                         "Exit elastic-training rendezvous when there are "
                         "agents to join the network-check rendezvous."
                     )
+                    self._check_round = 1
```
workingloong, line 833: What's happen if the first round check fails when `check_round=1`?

workingloong:
If exclude_straggler is not set, the check will pass even if the straggler exists. By the way, what's the result do you want if the straggler always exists?

PR #1387, opened 2024-12-12
Set default brain addr

sunjq1:

This PR enhances the code's robustness by setting `defaultBrainServiceAddr` when `optimizeMode` is `cluster`.

If `optimizeMode` is configure to `cluster` in the yaml file but `brainService` is not configured, the only difference from configuring `optimizeMode` to 'single-job' is the warning message in the master container logs: 'Fail to JobMetricCollector.collect_custom_data by 'NoneType' object has no attribute 'persist_metrics'.

No.

UT and training test.


workingloong (approved):
LGTM

PR #1388, opened 2024-12-12
fix node.is_released lost in process_event

majieyue:

the node is deleted but the node.is_released is not updated

the previous node.is_released is a deepcopy so node.is_released = true won't work
we need to add it in node.update_info

NO

UT


samplise (approved):
lgtm

PR #1407, opened 2024-12-26
check if pod is already running before relaunch

majieyue:

check pod is duplicated before relaunch another one

some k8s plugin can recover pod from starting failure. when a pod is deleted or modified from a pending/running state, the pod with the same rank and id may be already running successfully. so we should not relaunch another same rank pod

No

BVT UT

### DIFF
--- a/dlrover/python/master/node/dist_job_manager.py
+++ b/dlrover/python/master/node/dist_job_manager.py
@@ -790,6 +790,34 @@ class DistributedJobManager(JobManager):
             should_relaunch = self._should_relaunch(
                 cur_node, status_change_flow
             )
+
+            # Even _should_relaunch return True by state-machine check,
+            # we need another round of check
+            if should_relaunch:
+                selector = k8s_util.gen_k8s_label_selector_from_dict(
+                    self._get_pod_unique_labels(cur_node)
+                )
+                logger.info(
+                    f"Recheck relaunch node: {self._job_args.job_name}, "
+                    f"{cur_node.type}, {cur_node.rank_index}, {cur_node.id} "
+                    f"with label {selector}"
+                )
+                pods = self._k8s_client.list_namespaced_pod(selector)
+                if (
+                    pods
+                    and len(pods.items) > 0
+                    and any(
+                        pod.status.phase == NodeStatus.RUNNING
+                        and not pod.metadata.deletion_timestamp
+                        for pod in pods.items
+                    )
+                ):
+                    logger.info(
+                        f"Skip relaunch node {cur_node.id}({cur_node.name}) "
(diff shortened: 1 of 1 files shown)


dlrover/python/master/node/dist_job_manager.py
```diff
@@ -790,6 +790,34 @@ def _process_event(self, event: NodeEvent):
             should_relaunch = self._should_relaunch(
                 cur_node, status_change_flow
             )
+
+            # Even _should_relaunch return True by state-machine check,
+            # we need another round of check
+            if should_relaunch:
```
workingloong, line 796: Move these codes into the method `_should_relaunch` and add UT.

workingloong (commented):
Does the native k8s recover the pod from starting failure?

dlrover/python/master/node/dist_job_manager.py
```diff
@@ -790,6 +790,34 @@ def _process_event(self, event: NodeEvent):
...
+                selector = k8s_util.gen_k8s_label_selector_from_dict(
+                    self._get_pod_unique_labels(cur_node)
+                )
+                logger.info(
```
BalaBalaYi, line 800: This part has actually already been implemented (when generating the event before relaunch), so the issue probably isn't here.

BalaBalaYi (changes requested):
need more investigation

dlrover/python/master/node/dist_job_manager.py
majieyue: it is a k8s plugin going to open sourced

dlrover/python/master/node/dist_job_manager.py
majieyue: event type is MODIFIED and node status is RUNNING from log. so the condition in _process_event is not matched.
we didn't see the log either

dlrover/python/master/node/dist_job_manager.py
majieyue: ok

majieyue:
> Does the native k8s recover the pod from starting failure?

it is a k8s plugin going to open sourced

dlrover/python/master/node/dist_job_manager.py
BalaBalaYi: modified event with deletion timestamp will transfer to deleted event:
https://github.com/intelligent-machine-learning/dlrover/blob/5d070f49da6bde5389b0d282a1712ff6cf27d2b8/dlrover/python/master/watcher/k8s_watcher.py#L232

BalaBalaYi:
The issue is resolved by: https://github.com/intelligent-machine-learning/dlrover/pull/1408.

PR #1420, opened 2025-01-03
if node failed and exited, need relaunch on heartbeat timeout

majieyue:

change dead node skip to only succeeded exit node

if node failed and exited, but stuck into some post-mortem action such as coredump upload, we won't wait for it finishing, but relaunch as soon as the heartbeat timeout

No

UT


dlrover/python/master/node/dist_job_manager.py
```diff
@@ -510,7 +510,7 @@ def _get_dead_node_event(self, window_interval=600) -> List[NodeEvent]:
                     and node.start_time
                     and node.create_time
                     and node.status == NodeStatus.RUNNING
-                    and not node.is_exited_reported()
+                    and not node.is_succeeded_and_exited()
```
BalaBalaYi, line 513: Why make this change? Is the same.
```
  def is_exited_reported(self):
      return (
          self.reported_status == NodeEventType.SUCCEEDED_EXITED
          or self.reported_status == NodeEventType.FAILED_EXITED
      )
```

BalaBalaYi (approved):
LGTM

Issue #1366, opened 2024-11-29
Support HTTP for master-worker communication.


BalaBalaYi:
**Is your feature request related to a problem? Please describe.**
Considering the use of gRPC involves significant dependency issues, and there is a degree of uncertainty when using gRPC in multi-process and multi-threaded scenarios. Additionally, related issues might take a long time to be resolved. Currently, the gRPC interactions in DLRover are implemented at the control plane level, and even in large-scale scenarios, there are no extreme performance requirements. Therefore, there are plans to support an HTTP implementation as a replacement for the gRPC implementation.

**Describe the solution you'd like**
Support http for master-worker communication.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

BalaBalaYi:
https://github.com/intelligent-machine-learning/dlrover/pull/1429

PR #1429, opened 2025-01-08
Support http communication implement for DLRover Master and Agent.

BalaBalaYi:

1. Implement a http server based on tornado in master.
2. Adapt the implementation of the HTTP framework, including both the server and client, without modifying the core business logic, while also maintaining support for gRPC.

https://github.com/intelligent-machine-learning/dlrover/issues/1366

User can specify which service to use for communication between the master and agent by using the parameter ```--service_type=(http/grpc)```. For details, please refer: dlrover/docs/deployment/argument.md.

UT + training with grpc type service + training with http type service.


dlrover/python/common/http_server.py
```diff
@@ -0,0 +1,103 @@
...
+            self._serving_started = True
+
+            server_thread = threading.Thread(
+                target=self._start_server,
+                name=TornadoHTTPServer.SERVING_THREAD_NAME,
+            )
+            server_thread.start()
+
+            # wait 3s for sever start
+            time.sleep(3)
```
workingloong, line 81: Can we implement a function to check the server is ready not wait a fixed time.

workingloong:
It is very meaningful job. Can we remove the grpc in the following verison?

dlrover/python/common/http_server.py
BalaBalaYi: done

BalaBalaYi:
> It is very meaningful job. Can we remove the grpc in the following verison?

Maybe. We need more time and practice to verify the performance, reliability, flexibility, etc., of the HTTP approach.

majieyue (approved):
so far the code LGTM. but we need some follow up
1. perf report comparing grpc with http
2. security extensions in the future, e.g. basic authentication, crypt/decrypt etc. for now the message with auth is needed to prevent bot sending deprecated messages

samplise (approved):
lgtm

BalaBalaYi:
Note: The current implementation only indicates that the 'HTTP mode' is basically functional, but it does not imply that it is suitable for large-scale use in a production environment. We will continue to iterate and enhance the relevant parts until the functionality is stable and the performance meets expectations.

PR #1439, opened 2025-01-14
reduce check fault timeout and check straggler timeout to 1 sec

majieyue:
reduce the check fault and straggler nodes timeout to 1s, so we can make sure next nodecheck round the fault nodes and straggler nodes are synced up


BalaBalaYi (approved):
LGTM

PR #1447, opened 2025-01-21
Fix network check status report issue.

BalaBalaYi:

1. Use 'success' as flag to determine whether to use 'node-check failed' or 'node-check success' for reporting.
2. Use a defined error 'NodeCheckFailedError' to distinguish 'failed' and 'node-check failed' reporting.

1. Should return 'node-check success' if there is 1 round check which result is success.
2. Should not use 'failed exited' status to override 'node-check failed' status if the node is failed by node-check.

No.

Training with node check failed.

### DIFF
--- a/dlrover/python/elastic_agent/torch/training.py
+++ b/dlrover/python/elastic_agent/torch/training.py
@@ -143,6 +143,10 @@ class RendezvousOutSyncError(Exception):
     pass
 
 
+class NodeCheckFailedError(RuntimeError):
+    pass
+
+
 @dataclass
 class ElasticLaunchConfig(LaunchConfig):
     """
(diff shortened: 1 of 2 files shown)


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -1420,17 +1437,19 @@ def run(self, role: str = DEFAULT_ROLE) -> bool:
                 f"Network check time of round {i} is {elapsed_time}"
                 f" and succeed is {result}."
             )
+
+            success = success or result
```
samplise, line 1441: When round 1 is a success but round 2 is a failure, it still reports success as the final result. Why?

BalaBalaYi: The result reported here is the result status of the current node, not the result status of the group it participates in. The first time being successful and the second time failing indicates that it has been affected by others.

samplise (approved):
lgtm

PR #1450, opened 2025-01-22
Refactor failed node counter.

BalaBalaYi:

Use a dict to save fault node info in job context. 
Node check failed should also be counted as 'failed worker'.

Bug fix: some worker failed is not counted.

No.

Training with node check failed.


majieyue (approved):
lgtm

PR #1455, opened 2025-01-27
reduce rendezvous and sync port timeout to 5s

majieyue:

reduce rendezvous and sync port timeout to 5s

the rendezvous timeout is 15s which is too long, also the sync port timeout (20 sec)

NO

UT

### DIFF
--- a/dlrover/python/common/constants.py
+++ b/dlrover/python/common/constants.py
@@ -387,6 +387,12 @@ class JobConstant(object):
 
     TRAINING_AGENT_LOOP_DEFAULT_INTERVAL = 15
 
+    # sleep 5s before next rendezvous round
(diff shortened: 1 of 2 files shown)


dlrover/python/common/constants.py
```diff
@@ -387,6 +387,12 @@ class JobConstant(object):
 
     TRAINING_AGENT_LOOP_DEFAULT_INTERVAL = 15
```
workingloong, line 388: It is better to remove the constant if it is no longer used.

majieyue: it is still used somewhere else e.g. _initialize_workers

BalaBalaYi (approved):
LGTM.

PR #1459, opened 2025-02-03
The go master sets Pod owner reference as the ElasticJob instance.

workingloong:

The go master sets Pod owner reference as the ElasticJob instance.

The pod should be deleted once when the job is deleted.

No.

UT.


samplise (approved):
lgtm

PR #1461, opened 2025-02-05
add exception catch in _initialize_workers

majieyue:

add final exception handler in _initialize_workers

Due to unexpected corner-cases or bugs, which sometimes comes from torch, the _initialize_workers
may throw exceptions other than RendezvousOutSyncErro, which we need to handle gracefully.
Otherwise the dlrover agent would fail and exit

No

UT

### DIFF
--- a/dlrover/python/elastic_agent/torch/training.py
+++ b/dlrover/python/elastic_agent/torch/training.py
@@ -892,9 +892,9 @@ class ElasticTrainingAgent(LocalElasticAgent):
                 if err_cnt < max_errors:
                     stack_trace = traceback.format_exc()
                     logger.error(
-                        f"Unexpected exception in _initialize_workers: {e}"
+                        f"Unexpected exception in _initialize_workers: {e}\n"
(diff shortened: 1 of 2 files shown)


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -885,6 +887,18 @@ def _initialize_workers(self, worker_group):
                 time.sleep(JobConstant.TRAINING_AGENT_LOOP_DEFAULT_INTERVAL)
                 if time.time() - start_pending > pend_timeout:
                     raise TimeoutError("Timeout to wait for new nodes.")
+            except Exception as e:
+                err_cnt += 1
+                if err_cnt < max_errors:
+                    stack_trace = traceback.format_exc()
+                    logger.error(
```
BalaBalaYi, line 894: Merge the following two 'logger.error' into one part.

BalaBalaYi (approved):
LGTM

PR #1462, opened 2025-02-07
Rm master report event in training subprocess.

BalaBalaYi:

Use local event logging instead of reporting to master for training process when doing ckpt.

1. The event reporting of the training process can be done through other means (event exporter).
2. Avoid excessive event reporting, which can exhaust the performance of the master server and lead to other stability issues.

No

UT and training


samplise (approved):
lgtm

majieyue (approved):
lgtm

PR #1464, opened 2025-02-10
optimize the shared lock

majieyue:
Currently the acquire and release of shared lock needs the client opening a new connection, and in case client crashed without releasing locks, the lock is dead permanently. 

We introduce persistent connection in LocalSocketComm so the client with connection will be kept and reused, until the SharedLock is deleted. The SharedLock will use persistent connection so the server will detect if connection is alive and release the lock if the client with lock is broken.

introduce persistent mode in LocalSocketComm
SharedLock use persistent connection to the server

No

UT


dlrover/python/common/multi_process.py
```diff
@@ -239,11 +244,14 @@ def _sync(self):
...
-        client = _create_socket_client(self._socket_file)
+        if self._client is None:
+            self._client = _create_socket_client(self._socket_file)
         message = pickle.dumps(request)
-        _socket_send(client, message)
-        rcv_data = _socket_recv(client)
-        client.close()
+        _socket_send(self._client, message)
+        rcv_data = _socket_recv(self._client)
+        if not self._persist:
```
BalaBalaYi, line 252: Should we move ```client.close()``` into 'finally' block?

dlrover/python/common/multi_process.py
```diff
@@ -267,45 +275,81 @@ class SharedLock(LocalSocketComm):
             the lock.
```
BalaBalaYi, line 275: Annotations need to be updated.

dlrover/python/common/multi_process.py
```diff
@@ -267,45 +275,81 @@ class SharedLock(LocalSocketComm):
...
+                break
+            except Exception as e:
+                logger.error(f"SharedLock unexpected recv error: {e}")
+                response = SocketResponse()
+                response.status = ERROR_CODE
+
+            try:
+                send_data = pickle.dumps(response)
+                _socket_send(connection, send_data)
+            except Exception as e:
```
BalaBalaYi, line 326: Should we catch the exception here? The upstream caller won't know that an error occurred here if so.

dlrover/python/common/multi_process.py
```diff
@@ -351,10 +399,20 @@ def locked(self):
         else:
             request = SocketRequest(
                 method="locked",
+                id=self._id,
                 args={},
             )
             return self._request(request)
 
+    def close(self):
```
BalaBalaYi, line 407: Is this function used for ut only?

BalaBalaYi:
'The SharedLock will use persistent connection so the server will detect if connection is alive and release the lock if the client with lock is broken.' Actually i don't understand where this point is reflected. Could you explain it further with some practical examples?

dlrover/python/common/multi_process.py
majieyue: not really

dlrover/python/common/multi_process.py
majieyue: only close in non-persistent cases

dlrover/python/common/multi_process.py
majieyue: done

BalaBalaYi:
> 'The SharedLock will use persistent connection so the server will detect if connection is alive and release the lock if the client with lock is broken.' Actually i don't understand where this point is reflected. Could you explain it further with some practical examples?

Already discussed.

BalaBalaYi (approved):
LGTM

PR #1467, opened 2025-02-10
Further optimization implementation for the pre-check(diagnosis) process.

BalaBalaYi:

1. Add 'node_id', 'node_type' definition in ```NodeAction```.
2. Move the action consuming thread into 'master' instead of 'job-manager', and split the specific implementation into the 'master' and 'job-manager'. 
3. Optimize ```PreCheckOperator``` abstraction for better flexibility.
4. Add ```JobAbortionAction``` and the procedure for it.
5. Add ```SchedulingPreCheckOperator``` implementation.
6. Other compatibility adjustments.

1. Refactor some diagnosis action's definition and the action's execution to support:

    a. Master can relaunch specified node according to the action
    b. Master can abort job execution according to the action

2. Refactor input args to support user defined pre-check operators.
3. Pre-check procedure full implementation.

User can use 'pre_check_ops' to define the operators for the pre-check procedure. Details please refer to: [argument.md](https://github.com/BalaBalaYi/dlrover/blob/optimize_pre_check/docs/deployment/argument.md)

UT and Training under different circumstances.


dlrover/python/master/diagnosis/diagnosis_manager.py
```diff
@@ -128,16 +126,22 @@ def pre_check(self):
...
                     f"by {pre_check_op_name} "
                     f"with result: {current_op_result}, "
                     f"cost:{time.time()-current_start:.2f}ms. "
-                    f"Invoke action: {action}."
+                    f"Invoke action: {actions}."
                 )
-                self._job_context.set_pre_check_status(PreCheckStatus.FAIL)
+                if _dlrover_context.pre_check_bypass_enabled:
+                    logger.warning(
+                        "Set pre-check pass due to " "bypass is enabled."
```
BalaBalaYi, line 140: lint

dlrover/python/common/constants.py
```diff
@@ -325,7 +326,7 @@ class RendezvousName(object):
 
 
 class NodeErrorMessage(object):
-    NETWORKER_ERROR = "Network is breakdown"
+    NODE_CHECK_ERROR = "Node check has failure"
```
workingloong, line 329: "Node check has failure" -> "Node check not passed"

BalaBalaYi: "Node check failed"

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -1510,7 +1519,7 @@ def run(self, role: str = DEFAULT_ROLE) -> bool:
 
         if self._node_rank in fault_nodes:
             self._client.report_failures(
-                NodeErrorMessage.NETWORKER_ERROR,
```
majieyue, line 1513: the agent should block util the master finish prre-check, but I didn't find the code

dlrover/python/master/args.py
```diff
@@ -66,11 +118,21 @@ def add_params(parser):
...
+        "Pre training check will be disabled if parameter is empty.",
+    )
+    parser.add_argument(
+        "--pre_check_bypass",
+        "--pre-check-bypass",
+        default=DefaultValues.PRE_CHECK_BYPASS,
+        type=parse_tuple2_dict,
+        help="Will set pass result although pre-check failed if set bypass, "
+        "format: {(${module_name}, ${class_name}): ${boolean}, ...}.",
     )
```
majieyue, line 137: it is so hard for users to use class name as parameters, can we just make it easier

dlrover/python/common/global_context.py
```diff
@@ -67,7 +69,13 @@ class DefaultValues(object):
...
     NPU_NUM_PER_NODE = 16
     MAX_METRIC_REC = 30
-    PRE_CHECK_ENABLED = True
+    PRE_CHECK_OPS: List[Tuple] = [
+        (
+            "dlrover.python.master.diagnosis.precheck_operator",
+            "NoPreCheckOperator",
+        )
+    ]
+    PRE_CHECK_BYPASS: Dict[Tuple, bool] = {}
```
majieyue, line 78: why not use PRE_CHECK_ENABLE/DISABLE only. leave the details in the code and it is friendly to users.

dlrover/python/master/dist_master.py
```diff
@@ -203,9 +210,47 @@ def prepare(self):
         if self.job_manager:
             self.job_manager.start()
 
+        threading.Thread(
+            target=self._diagnose_job,
+            name="job_diagnosing",
+            daemon=True,
+        ).start()
+
```
majieyue, line 218: I think dist_master should wait for all agents running before starting a pre_check, but didn't find it

dlrover/python/common/global_context.py
BalaBalaYi: Actually, as you mentioned, the first version was implemented in that way, but the reasons for changing it to the current format are as follows:

1. The pre-check operator is a serial expression, and as daily development progresses, the corresponding operators will gradually increase. The bypass logic must support targeting a specific operator (it cannot only serve as a global switch).
2. This parameter is a high-level parameter and is not intended for regular users. It is typically configured by developers who have worked on the operator, to validate the usability of newly added operators. Once the operator functionality stabilizes, the configuration becomes relatively fixed and does not require frequent changes.
3. If a user needs to disable all operator checks, configuring pre_check_op=[] is an easier implementation.

In summary, this parameter's implementation emphasizes flexibility over usability. Of course, if there are easier implementation options that satisfy the above conditions, we welcome any suggestions.

dlrover/python/master/dist_master.py
BalaBalaYi: The logic u mentioned is in ```SchedulingPreCheckOperator```.
BTW,  the pre-check process doesn't necessarily have to start after scheduling; in the future, it may involve some additional checks that need to be performed before (or during) scheduling.

dlrover/python/elastic_agent/torch/training.py
BalaBalaYi: Here:
https://github.com/BalaBalaYi/dlrover/blob/14284a4c1b6fde6a4a2975520a339993a7a3f6ee/dlrover/trainer/torch/elastic_run.py#L269

dlrover/python/master/args.py
BalaBalaYi: Please refer to the another comment.

dlrover/python/common/constants.py
```diff
@@ -325,7 +332,7 @@ class RendezvousName(object):
 
 
 class NodeErrorMessage(object):
-    NETWORKER_ERROR = "Network is breakdown"
+    NODE_CHECK_FAIL = "Node check failed"
```
samplise, line 335: `NODE_CHECK_FAILED`

dlrover/python/common/global_context.py
samplise: Since `PRE_CEHCK_OPS` and `PRE_CHECK_BYPASS` both include the complete operator package and name, it looks a little redundant here. How about only have `PRE_CHECK_BYPASS` and rename it to be `PRE_CHECK_OPS`?

dlrover/python/common/global_context.py
```diff
@@ -208,3 +218,29 @@ def print_config(self):
...
+            and len(self.get_pre_check_operators()) > 0
+        )
+
+    def get_pre_check_operators(self):
+        result_ops = []
+        if not self.pre_check_operators:
+            return result_ops
+
+        for op_desc in self.pre_check_operators:
+            try:
```
samplise, line 234: Why not check bypass here?

dlrover/python/diagnosis/common/diagnosis_action.py
```diff
@@ -214,6 +234,45 @@ def node_status(self):
...
+            f"timestamp:{self._timestamp};"
+            f"expired_time_period:{self._expired_time_period};"
+            f"node_id:{self._node_id};"
+            f"node_type:{self._node_type};"
+            f"node_status:{self._node_status};"
+            f"reason:{self._reason}"
+        )
+
+
+class JobAbortionAction(DiagnosisAction):
```
samplise, line 250: How about defining a job action and making AbortAction as a sub type of it?

majieyue (approved):
lgtm

dlrover/python/common/constants.py
BalaBalaYi: Rm ```NodeErrorMessage``` and use ```NodeEventType``` instead.

dlrover/python/common/global_context.py
BalaBalaYi: Also for above question: `Since PRE_CEHCK_OPS and PRE_CHECK_BYPASS both include the complete operator package and name, it looks a little redundant here. How about only have PRE_CHECK_BYPASS and rename it to be PRE_CHECK_OPS?`

Good comment. A new way:

1. use 'PRE_CHECK_OPS' only
2. 'PRE_CHECK_OPS' 's format is {tuple2:bool}

- not use operator: no configuration at all  e.g. {}
- use operator but bypass: module and class configuration with 'False' value  e.g. {(m1, c1): False}
- use operator without bypass: module and class configuration with 'True' value  e.g. {(m1, c1): True}

dlrover/python/diagnosis/common/diagnosis_action.py
BalaBalaYi: I’ve already tried that, but I found it somewhat redundant. When you define ```JobAction```, you’ll notice that there isn’t much unique to express under it—it’s essentially the same as the Action base class.

dlrover/python/common/global_context.py
BalaBalaYi: > Since `PRE_CEHCK_OPS` and `PRE_CHECK_BYPASS` both include the complete operator package and name, it looks a little redundant here. How about only have `PRE_CHECK_BYPASS` and rename it to be `PRE_CHECK_OPS`?

Good suggestion. Already done.

dlrover/python/master/diagnosis/diagnosis_manager.py
```diff
@@ -53,103 +53,170 @@
...
         )
 
-        for pre_check_op in pre_check_ops:
-            current_start = time.time()
-            current_op_result = None
-            pre_check_op_name = pre_check_op.__class__.__name__
+        round = 0
+        pre_check_finish = False
+
+        while True:
```
samplise, line 107: 1. Add comments to explain the key idea here. For example, when an operator fails, it will bypass/retry/execute failure action. 
2. When one operator fails, why retry all precheck operators ?

dlrover/python/master/diagnosis/precheck_operator.py
```diff
@@ -44,39 +59,269 @@ def get_retry_interval_secs(cls) -> int:
...
+                pending_result = self.check_ps_job_pending(
+                    ps_nodes + worker_nodes, timeout, strategy
+                )
+            else:
+                msg = f"Skip {self.__class__.__name__} for {job_type}."
+                logger.warning(msg)
+                return PreCheckResult(result_msg=msg)
+
+            if pending_result[0]:
+                # has pending
```
samplise, line 300: `has pending` -> `job is pending`

dlrover/python/master/diagnosis/precheck_operator.py
```diff
@@ -44,39 +59,269 @@ def get_retry_interval_secs(cls) -> int:
...
+                if pending_result[1]:
+                    # has pending timeout
```
samplise, line 302: `job is pending timeout`

dlrover/python/master/diagnosis/precheck_operator.py
```diff
@@ -44,39 +59,269 @@ def get_retry_interval_secs(cls) -> int:
...
+                    return PreCheckResult(
+                        result=1,
+                        result_msg=SchedulingPreCheckOperator.PENDING_TIMEOUT_MSG,  # noqa: E501
+                        abnormal_nodes=[pending_result[1]],
+                    )
+                else:
+                    # has pending wait
+                    # continue waiting next round checking
```
samplise, line 310: remove `has pending wait`, `continue waiting next round checking` -> `wait for a while and try checking again`

dlrover/python/master/diagnosis/precheck_operator.py
BalaBalaYi: Is node pending, we don't have the concept of 'job pending'.

dlrover/python/master/diagnosis/precheck_operator.py
BalaBalaYi: same

dlrover/python/master/diagnosis/precheck_operator.py
BalaBalaYi: done

dlrover/python/master/diagnosis/diagnosis_manager.py
BalaBalaYi: 1. Done.
2. Yes. Here's a simple example: if the scheduling check is performed first, followed by the node check, and it is found that some nodes need to be replaced, then after triggering the relaunch logic, the scheduling check needs to be performed again to ensure that the relaunched nodes can be scheduled properly.

PR #1476, opened 2025-02-24
fix signal hanlder: don't unregister if origin signal handler is SIG_IGN

liyzcj:

fix signal hanlder: don't unregister if origin signal handler is SIG_IGN

if unregister for all signals, signals after SIG_IGN will be lost.

No

UT


majieyue (approved):
lgtm

PR #1483, opened 2025-02-27
Fix and enhance master connection issue.

BalaBalaYi:

1. Refactor ```check_to_use_dlrover_run```.
2. Move 'retry' annotation into ```function_util```.

For typical production scenarios, the use of local-master is not recommended. Therefore, corresponding switches have been added to further optimize the initialization logic in actual production scenarios.

For now the implementation of worker-master connection scenarios: 

- standalone mode:
  - dlrover-run with  local-master
  - torchrun without master

- distributed mode:
  - dlrover-run with  distributed-master

- other cases: will raise exception directly

No(still using --standalone param in dlrover-run command for certain strategy).

UT and Trainning.

### DIFF
--- a/dlrover/python/common/event/reporter.py
+++ b/dlrover/python/common/event/reporter.py
@@ -10,12 +10,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import importlib
 import json
 from datetime import datetime
 from typing import Dict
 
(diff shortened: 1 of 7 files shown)


majieyue:
I think it is better to constrain when use dlrover master or local master. The two cases should not be compatible with each other.
If job is started by dlrover master, it should rollback to local master even if dlrover master failed. Otherwise if job started with a local master param, it should only run within a node, and I think the 2nd case, a dlrover-run command would be enought

majieyue (approved):
lgtm

PR #1486, opened 2025-03-04
use RdzvTimeoutError to replace builtin TimeoutError

majieyue:

We add an exception handler in _initialize_workers and when TimeoutError raised, it will retry util max_errors reached.
But during pending or rendezvous timeout, it is better to raise it out instead of retrying, to failover fastly

To let pending timeout and rendezvous timeout to fail-fast, instead of retry many times

No

UT

### DIFF
--- a/dlrover/python/elastic_agent/torch/training.py
+++ b/dlrover/python/elastic_agent/torch/training.py
@@ -164,6 +164,10 @@ class RendezvousTimeoutError(RuntimeError):
     pass
 
 
+class StopWorkerTimeoutError(RuntimeError):
+    pass
+
+
 @dataclass
 class ElasticLaunchConfig(LaunchConfig):
     """
@@ -917,7 +921,11 @@ class ElasticTrainingAgent(LocalElasticAgent):
                     raise RendezvousTimeoutError(
                         "Timeout to wait for new nodes."
                     )
-            except (NodeCheckFailedError, RendezvousTimeoutError) as e:
+            except (
+                NodeCheckFailedError,
+                RendezvousTimeoutError,
+                StopWorkerTimeoutError,
+            ) as e:
                 raise e
             except Exception as e:
                 err_cnt += 1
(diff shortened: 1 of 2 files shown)


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -160,6 +160,10 @@ class NodeCheckFailedError(RuntimeError):
     pass
 
 
+class RdzvTimeoutError(RuntimeError):
```
BalaBalaYi, line 163: The naming convention should be consistent with ```RendezvousOutSyncError ```

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -946,14 +952,14 @@ def _stop_workers(
                     super()._stop_workers(worker_group, is_restart)
 
             signal.alarm(0)
-        except TimeoutError as te:
+        except RdzvTimeoutError as te:
```
BalaBalaYi, line 955: Should not use 'RdzvTimeoutError' instead of 'TimeoutError' here. The implement here is not correlated with rendezvous.

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -946,14 +952,14 @@ def _stop_workers(
...
+        except RendezvousTimeoutError as te:
             logger.error(str(te))
-            raise
+            raise te
         finally:
             signal.alarm(0)
 
     def _stop_timeout_handler(self, signum, frame):
-        raise TimeoutError("Timed out waiting for stopping workers.")
+        raise RendezvousTimeoutError("Timed out waiting for stopping workers.")
```
BalaBalaYi, line 962: The issue still persists: the timeout here for 'stop_workers' is unrelated to the rendezvous (rdzv).

BalaBalaYi (approved):
LGTM

PR #1497, opened 2025-03-11
Fix insufficient judgement.

BalaBalaYi:

Add ```len(available_nodes) > 0``` as a condition. 
Because if there are no 'available_nodes 'at the moment, it implies one of two scenarios:

1. All nodes have succeeded.
2. All nodes have failed (and exited at the same moment).

In the case of the first scenario, directly ignoring it aligns with expectations. 

For the second scenario, if the failure causes an exit, there will inevitably be fault tolerance (as long as the master hasn't exited), meaning this logical judgment process will be entered again.

Bug fix.

NO

UT


samplise (approved):
lgtm

PR #1499, opened 2025-03-18
Add a new debug parameter 'task_process_timeout' arg

BalaBalaYi:

1. new 'task_process_timeout' argument
2. args parser optimized

Introducing a new debug parameter.  

This new parameter will not be directly exposed for now (it is intended for internal debugging purposes only).  

UT and Training with this argument specified.


samplise (approved):
lgtm

PR #1505, opened 2025-03-25
Fix pre-check issue.

BalaBalaYi:

Skip pre-check if the result already set to pass.

Fix pre-check issue.

No.

UT and training job with master failover.


majieyue (approved):
lgtm

PR #1508, opened 2025-03-27
if succeeded and exited, delete this pod if timeout

majieyue:

In pod scaler list operation, if the pod is still in job nodes and its reported_status is SUCCEEDED_AND_EXITED, which means
the agent has sent master its exiting message, but the pod is still running. In case of that scaler will delete this pod

If agent exited but the pod is still running, dlrover master can not remove this node from its nodes context. So we need another way of removing a "SUCCEEDED_AND_EXITED" RUNNING POD
In pod scaler list operation, if the pod is still in job nodes and its reported_status is SUCCEEDED_AND_EXITED, which means
the agent has sent master its exiting message, but the pod is still running. In case of that scaler will delete this pod

No

UT


dlrover/python/master/watcher/k8s_watcher.py
```diff
@@ -268,6 +273,23 @@ def list(self) -> List[Node]:
...
+                target_node = job_ctx.job_node(node_type, node_id)
+                now = int(datetime.now().timestamp())
+                if target_node:
+                    status, ts = target_node.reported_status
+                    if (
+                        status == NodeEventType.SUCCEEDED_EXITED
+                        and now - ts > 600
+                    ):
+                        logger.info(
+                            f"Target pod: {pod_name} {pod_type} {pod_id} "
```
BalaBalaYi, line 288: f"Delete target pod {pod_name} for timeout due to succeeded status after 600s."

BalaBalaYi (approved):

PR #1521, opened 2025-04-14
fix pickle CVE of executing illegal instructions or commands

majieyue:

use dlrover_pickle to replace pickle to secure illegal unpickling

pickle is unsecure as we known for a long time. since dlrover also uses pickle to deserialize messages, we need to restrict the module that can be unpickled

No

UT


BalaBalaYi (approved):
LGTM

PR #1524, opened 2025-04-23
Fix none type error in pre-check.

BalaBalaYi:

To avoid create time is none when sorting.

Bug fix.

NO

Training job


majieyue (approved):
lgtm

PR #1525, opened 2025-04-24
reimplement metric diagnosis, combine step with tensor metrics

majieyue:
We enable step collector for atorch trainer, so the starting and ending of each step will be reported to master.
If step has begun but not finished before timeout, it might be hang.
We are also continuing to collect key metrics such as tensor util, if tensor drop to zero for a period of time and the step 
is also stuck, the job can be regarded as hang job and we will restart the processes to retry

To improve training availability

No

UT


dlrover/python/common/event/context.py
```diff
@@ -68,6 +70,23 @@ def get_first_step_event(self):
             key = keys[0]
             return self._step_events[key]
 
+    def last_steps_avg_time(self, last_steps):
```
BalaBalaYi, line 73: steps -> step

dlrover/python/common/event/context.py
```diff
@@ -77,80 +96,115 @@ def add_step_event(self, event: AtorchEvent):
...
+                step_event = AtorchStepEvent(
+                    event.name, TrainEventState.TRAIN_EVT_BEGIN, event.step
+                )
+                step_event.begin_timestamp = event.timestamp
+                step_event.localtime = int(datetime.now().timestamp())
+                self._step_events[event.timestamp] = step_event
+
+            elif event.type == EventTypeName.END:
+                if not len(keys):
+                    logger.error(f"invalid ckpt step without BEGIN: {event}")
```
BalaBalaYi, line 183: replace with warning

dlrover/python/common/event/context.py
```diff
@@ -77,80 +96,115 @@ def add_step_event(self, event: AtorchEvent):
...
+            if event.type == EventTypeName.BEGIN:
+                if len(keys) > 0:
+                    last_event = self._step_events[keys[-1]]
+                    if last_event.event_state != TrainEventState.TRAIN_EVT_END:
+                        logger.error(
+                            f"invalid ckpt step: {last_event}, {event}"
+                        )
+                        return
+                    if event.timestamp < last_event.end_timestamp:
+                        logger.error(
```
BalaBalaYi, line 169: replace with warning

dlrover/python/common/event/context.py
```diff
@@ -77,80 +96,115 @@ def add_step_event(self, event: AtorchEvent):
...
+                step_event = self._step_events[last_key]
+                if (
+                    step_event.step != event.step
+                    or step_event.event_state
+                    != TrainEventState.TRAIN_EVT_BEGIN
+                ):
+                    logger.error(f"invalid ckpt step: {step_event}, {event}")
+                    return
+                if step_event.begin_timestamp > event.timestamp:
+                    logger.error(
```
BalaBalaYi, line 195: There are too many error logs of this type. There should be a unified event validation process here, where validation failures are logged as warnings. 
BTW, currently only event logs that impact core processes use the error level.

dlrover/python/common/event/context.py
```diff
@@ -77,80 +96,115 @@ def add_step_event(self, event: AtorchEvent):
...
-            elif event.type is EventTypeName.END:
-                last_step.finish_ckpt(event.timestamp)
-                last_step.localtime = int(datetime.now().timestamp())
+                step_event.end_timestamp = event.timestamp
+                step_event.step_time = (
+                    step_event.end_timestamp - step_event.begin_timestamp
+                )
+                step_event.step = event.step
+                step_event.event_state = TrainEventState.TRAIN_EVT_END
+                step_event.localtime = int(datetime.now().timestamp())
```
BalaBalaYi, line 206: int(datetime.now().timestamp()) -> int(time.time())
Since we are already using Unix timestamps here, there's no need to add an extra conversion through datetime.

dlrover/python/master/diagnosis/diagnosis_master.py
```diff
@@ -407,23 +422,26 @@ def check_tensor_drop_zero(self, duration):
...
-            if (
-                _dlrover_context.hang_downtime
-                < DefaultValues.MIN_HANG_DOWNTIME
-            ):
-                hang_downtime = DefaultValues.MIN_HANG_DOWNTIME
-            else:
-                hang_downtime = _dlrover_context.hang_downtime
+            if self._is_observing_paused:
+                logger.info(
+                    f"Pause _metric_diagnose thread due to "
```
BalaBalaYi, line 437: Pause _metric_diagnose thread due to paused status

dlrover/python/master/diagnosis/diagnosis_master.py
```diff
@@ -437,12 +455,17 @@ def _diagnose_metrics(self):
                 )
 
                 if _dlrover_context.hang_detection == 2:
-                    self._job_context.enqueue_diagnosis_action(
-                        NodeAction(
-                            action_type=DiagnosisActionType.RESTART_WORKER,
-                            instance=DiagnosisConstant.ANY_INSTANCE,
+                    if step_hang is True:
+                        logger.info("Restart worker-0 all processes")
```
BalaBalaYi, line 459: why only worker-0 need restart?

dlrover/python/master/servicer.py
```diff
@@ -300,6 +300,11 @@ def _join_rendezvous(self, request: comm.JoinRendezvousRequest):
                 RendezvousName.ELASTIC_TRAINING
             ]
             training_manager.clear_waiting_nodes()
+
+        # Pause hang diagnosis during rendezvous
+        if node_rank == 0:
```
BalaBalaYi, line 305: This implementation should not be placed here. It should be directly implemented under `rdzv_manager`.

dlrover/python/master/servicer.py
```diff
@@ -334,6 +339,9 @@ def _get_comm_world(self, request: comm.CommWorldRequest):
             rdzv_round = rdzv_manager.get_rdzv_round()
             metrics = {CustomMetricKeys.RDZV_ROUND: rdzv_round}
             self._job_metric_collector.collect_custom_data(metrics)
+            # Finish elastic training rendezvous so we continue diagnosis
+            self._diagnosis_manager.continue_observing()
```
BalaBalaYi, line 343: same this part

BalaBalaYi (changes requested):

PR #1527, opened 2025-04-26
support user-defined monitor in diagnosis

majieyue:

User can implement its own metric monitor and add it by diagnosis_manager.new_metric_monitor

More scalability

No

UT


BalaBalaYi (approved):
LGTM

PR #1533, opened 2025-05-06
Optimize task-manager when master exit.

BalaBalaYi:

Return true when dataset is empty when invoke 'finished'.

To avoid task-manager's warning(has unfinished task) when task-manager is not used.

NO

UT


majieyue (approved):
lgtm

PR #1536, opened 2025-05-09
Fix job manager endlessly waiting when pod reported SUCCEEDED_EXITED but DELETED when terminating timeout

majieyue:

If pod has reported SUCCEEDED_EXITED, but didn't terminated successfully, We need to tag it as Succeeded after all.

If worker finished but pod didn't exit after 10min, e.g. stuck in TerminatingPOD, some postmordem work hang etc., dlrover will DELETE the pod. So if the pod has reported SUCCEEDED_EXITED, the status is DELETED, and relaunch_count may be 0 because dlrover will not relaunch any pod in JOB_STOPPING stage. Meanwhile, although SUCCEEDED_EXITED reported, the pod staus is not "Succeeded" but "Deleted", so that job_manager will return True in pend_without_workers() and continue to wait endlessly

NO

UT and BVT

### DIFF
--- a/dlrover/python/master/node/worker.py
+++ b/dlrover/python/master/node/worker.py
@@ -280,7 +280,12 @@ class WorkerManager(TrainingNodeManager):
         return plan
 
     def has_exited_worker(self):
-        """Check whether there is exited worker except evicted workers."""
+        """Check whether there is exited worker except evicted workers.
+
+        If the worker has reported SUCCEEDED_EXITED, but been deleted
+        by dlrover finally, the status will be DELETED instead of SUCCEEDED
+        In such cases the worker should also be regard as exited worker
+        """
         nodes = self._get_nodes()
         for worker in nodes.values():
             if (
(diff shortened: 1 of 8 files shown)


dlrover/python/master/node/worker.py
```diff
@@ -285,7 +286,16 @@ def has_exited_worker(self):
             if (
                 worker.exit_reason == NodeExitReason.FATAL_ERROR
                 or worker.status == NodeStatus.SUCCEEDED
+                or (
+                    worker.status == NodeStatus.DELETED
+                    and worker.reported_status[0]
```
BalaBalaYi, line 291: Better to add a method under the `Node` class that only exposes the reported status(status without timestamp), to avoid the widespread use of `tuple[magic number]`.

dlrover/python/master/node/worker.py
```diff
@@ -285,7 +286,16 @@ def has_exited_worker(self):
             if (
                 worker.exit_reason == NodeExitReason.FATAL_ERROR
                 or worker.status == NodeStatus.SUCCEEDED
+                or (
```
BalaBalaYi, line 289: Also please add the reason described in the PR as a comment here.

BalaBalaYi (approved):

PR #1544, opened 2025-05-15
fix case: has node relaunched after master restart

BalaBalaYi:

According to the rank comparison, confirm whether the init status and non-released node has already been replaced by a new node.  

Bug fix.

NO

UT


majieyue (approved):
lgtm

PR #1550, opened 2025-05-20
Fix node exit reason.

BalaBalaYi:

1. Will extract reason from pod spec only when pos is in terminal status.
2. Fix wrong exit reason 'unknown error' for succeeded pod.

Bug fix

NO

UT and Training.


majieyue (approved):
lgtm

PR #1553, opened 2025-05-23
Fix mismatch for ElasticDistributedSampler when scaling.

zhengchenyu:

Fix mismatch for ElasticDistributedSampler when scaling.

When I scale down 4\*2 replicas to 3\*2 replica, I found exception:

```
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/dlrover/trainer/torch/elastic/sampler.py", line 112, in __iter__
[rank0]:     assert len(indices) == self.num_samples
```

After debugging, I found that when training is resumed from epoch i, this problem will occur at epoch i+1.
Reason:
When epoch=i, when training is resumed, dataset len = 60000, completed num = 56800, remaining_samples = 3200. 3200 cannot be divided by 6. So total_size is reset to 56800+3204=60004. Because 60004 cannot be divided by 6, it is reset to 60006. Note: **total_size is expanded twice**
At epoch=i+1, training will be restarted, and num_samples will be reset to dataset len / 6 = 10000. But total_size is still 60006, so len(indices)=10001. So total_size should be reinitialized when starting epoch, instead of using the result of the previous epoch.

But the root cause may have something to do with #1339. In #1339, we expanded total_size twice, then will result to mismatch. We don't have to adjust total_size.

NO.

Fix and  add unit-test, real job test.


workingloong (approved):
Good job!

PR #1562, opened 2025-05-30
Disable relaunch if node already relaunched.

BalaBalaYi:

Add a new exit reason 'relaunched' to tag the node who already relauched.

Ad title described.

NO

UT and Training with case: 1) node failover 2) than master failover


samplise (approved):
lgtm

workingloong:
The master will not relaunch the node if the relaunch count == the max relaunched count.

Issue #1529, opened 2025-04-30
Use megatron's ckpt impl directly for megatron version > 0.10


BalaBalaYi:
**Is your feature request related to a problem? Please describe.**
It is now recommended to directly use the checkpoint implementation from the latest version of Megatron. The dlrover's integration with Megatron's flash checkpoint implementation is based on version 0.6 (which is outdated), whereas the latest version of Megatron's implementation is relatively more mature.  

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1585, opened 2025-07-08
Avoid save_to_memory running too fast then causing save_to_storage to fail.

zhengchenyu:

If there is a save_to_storage operation before, save_to_men needs to wait for the last save_to_storage to complete.

If the frequency of save_to_memory is high enough that the next save_to_memory is triggered before save_to_stoarge is completed, save_to_stoarge will fail. The error is as follows:
```
[2025-07-08 11:24:07,066] [ERROR] [ckpt_saver.py:702:_save_shard] The step 100 in event is no equal to step 101 in memory.
```

Since it is recommended to use megatron checkpoint directly in #1529, the modifications related to megatron checkpoint are ignored here.

No

unittest, real ddp job.


dlrover/trainer/torch/flash_checkpoint/engine.py
```diff
@@ -204,8 +204,16 @@ def __init__(
                 name=CheckpointSharedObjPrefix.SAVE_STEP_QNAME + str(0),
                 create=False,
             )
+            self._notify_queue = SharedQueue(
+                name=CheckpointSharedObjPrefix.SAVE_STEP_QNAME
```
BalaBalaYi, line 208: implement this magic concatenation as a method of `ckpt_saver`

dlrover/trainer/torch/flash_checkpoint/fsdp_engine.py
```diff
@@ -487,6 +487,11 @@ def save_to_memory(self, step, state_dict, paths: Dict[str, str]):
         if self._local_rank != self.local_shard_id:
             return False
 
+        if self._checkpoint_event_step > 0:
```
BalaBalaYi, line 490: move this part into parent class(engine)

dlrover/trainer/torch/flash_checkpoint/full_ckpt_engine.py
```diff
@@ -112,6 +112,10 @@ def save_to_memory(self, step, state_dict, paths: Dict[str, str]):
                 ["model_states", "optim_states"] of the state dict and
                 the value is the path of storage to save.
         """
+        if self._checkpoint_event_step > 0:
+            notify_event = self._notify_queue.get()
+            assert notify_event.step == self._checkpoint_event_step
```
BalaBalaYi, line 117: what's ur intention if assert is false? 
In my personal opinion, this situation should be handled softly, such as skipping this step and providing a log explanation.  So the previous implementation that directly failed didn't have too many issues. At most, it was just not easy to understand why it was 'not equal' the log express.

dlrover/trainer/torch/flash_checkpoint/deepspeed_engine.py
```diff
@@ -120,6 +124,7 @@ def save_to_storage(self, step, state_dict, paths):
         if self._local_rank == 0 and success:
             event = CheckpointEvent(type=CheckpointEventType.SAVE, step=step)
             self._event_queue.put(event)
+            self._checkpoint_event_step = step
         if success:
             self.latest_step = step
```
qijiale76, line 129: All ranks should expect a notify to drain their local shard queue
```
        if self._local_rank == 0 and success:
            event = CheckpointEvent(type=CheckpointEventType.SAVE, step=step)
            self._event_queue.put(event)
        if success:
            self._checkpoint_event_step = step
            self.latest_step = step
```

PR #1590, opened 2025-07-18
Disable cpu hang detection when using gpu.

BalaBalaYi:

Skip hang value updating if gpu > 0.

The implementation for determining current CPU usage is not rigorous under GPU scenarios. Temporarily disabled. 

NO

UT


majieyue (approved):
lgtm

PR #1591, opened 2025-07-21
Fix tornado importing in servicer.

BalaBalaYi:

Both 'HttpMasterServicer' and 'HttpMasterHandler' need to be in the try-except.

Avoid import error(no tornado) when using ray.

NO

UT and Training job.


workingloong (approved):
LGTM

PR #1597, opened 2025-07-30
Support relaunch with node group

majieyue:

1. The process_node_list will update node the group information from k8s
2. The relaunch_node will update node group info into api server requests yaml

During node relaunch, the pod meta should include the original node group information

No

UT

### DIFF
--- a/dlrover/python/master/node/dist_job_manager.py
+++ b/dlrover/python/master/node/dist_job_manager.py
@@ -621,6 +621,16 @@ class DistributedJobManager(JobManager):
                     new_node = copy.deepcopy(node)
                     self._job_context.update_job_node(new_node)
 
+                # update node group info if necessary
+                if (
+                    node_type == NodeType.WORKER
+                    and node_id in job_nodes[node_type]
+                    and job_nodes[node_type][node_id].group != node.group
+                ):
+                    job_nodes[node_type][node_id].group = node.group
+                    job_nodes[node_type][node_id].group_size = node.group_size
+                    job_nodes[node_type][node_id].group_id = node.group_id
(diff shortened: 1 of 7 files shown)


dlrover/python/master/node/dist_job_manager.py
```diff
@@ -621,6 +621,16 @@ def _process_list_nodes(self, nodes: List[Node]):
                     new_node = copy.deepcopy(node)
                     self._job_context.update_job_node(new_node)
 
+                # update node group info if necessary
+                if (
+                    node_type == NodeType.WORKER
+                    and node_id in job_nodes[node_type]
+                    and job_nodes[node_type][node_id].group != node.group
```
BalaBalaYi, line 628: why need this condition: 'job_nodes[node_type][node_id].group != node.group' ?

dlrover/python/master/watcher/k8s_watcher.py
```diff
@@ -108,16 +108,32 @@ def _convert_pod_yaml_to_node(pod):
     node_group = None
     node_group_size = None
     node_group_id = None
+
+    logger.debug(
```
BalaBalaYi: keep 1 debug logging here

dlrover/python/master/node/dist_job_manager.py
majieyue: The scheduler will update pod labels after pod has been initialized and the event has already been handled by dlrover. So at first dlrover received pod initialization event and create node WITHOUT group info, before calling list pods api and fetched pods list with group info. In such cases, it will update the node

PR #1600, opened 2025-08-01
Update ci.

BalaBalaYi:

1. Use trusted publisher to publish whl.
2. No need to install extra'ray' for master image. Because master image is used for k8s only.

Update ci github actions.

no

workflow tested


samplise (approved):
lgtm

PR #1603, opened 2025-08-07
[DO NOT MERGE] Refactor relaunch node creation.

BalaBalaYi:

Unify the generation of the relaunch node object:
1) which attributes are inherited? 
2) which attributes are reset or updated
into the method: `generate_relaunch_node`.  

Optimize the relaunch node generation to avoid the risk of incorrect object attribute settings that could lead to other exceptions (which have occurred before).  

No

UT


dlrover/python/master/node/training_node.py
```diff
@@ -252,23 +251,11 @@ def relaunch_node(self, node: Node, remove_exited_node=False):
...
-            relaunch_node = node.get_relaunch_node_info(new_id)
+            new_name = self._new_node_name_fn(node.type, new_id)
+            relaunch_node = node.generate_relaunch_node(new_id, new_name)
             self._update_node(relaunch_node)
-        logger.info("Relaunch node %s to %s", node.name, new_id)
-        plan.launch_nodes.append(
-            Node(
-                node.type,
-                new_id,
-                copy.deepcopy(relaunch_node.config_resource),
```
BalaBalaYi, line 262: Please help confirm why it was necessary to use 'deepcopy' here before?  @workingloong

workingloong: deepcopy is more safe to avoid the  resource modification if the relaunch_node's resource is modified.

BalaBalaYi: > deepcopy is more safe to avoid the resource modification if the relaunch_node's resource is modified.

But I couldn't find any implementation of `resource` would be modified after the `_relaunch_node` method is called. The only i found is that the modification of resources due to OOM (Out Of Memory) issues occurs earlier in the `should_relaunch` method.
Also a `deepcopy` has already been performed in the `get_relaunch_node_info` method of the node.

BalaBalaYi: @samplise please have a look on this point

samplise: @workingloong I think it should be safe here. The `relaunch_node` is already a copy of the original node. There is not way to touch it from other places.

way-zer (approved):

PR #1610, opened 2025-08-13
drop out relaunched node group_id when group failover happened

majieyue:

We should not pass on node group id when group failover happened, since the scheduler will assign a new one

Make scheduler easy to differentiate it is a node group FO, or just a node FO in the same group

NO

UT


workingloong (approved):
LGTM

PR #1617, opened 2025-08-20
Handle empty opt plan

samplise:

Handle the case that opt plan is None.

Avoid errors when opt plan is None.

No.

Unit test

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -25,6 +25,6 @@ coverage:
         removed_code_behavior: fully_covered_patch
     project:
       default:
-        target: 80%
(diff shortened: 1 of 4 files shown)


BalaBalaYi (approved):

codecov.yml
```diff
@@ -24,6 +25,6 @@ coverage:
         removed_code_behavior: fully_covered_patch
     project:
       default:
-        target: 83%
+        target: 80%
```
BalaBalaYi:
```suggestion
        target: 81%
```

PR #1619, opened 2025-08-25
[Unified] Refactor Builder api

way-zer:

Update Builder api to align with internal changes. Changes including:
* Use pydantic to validate, when creating DLJob/RLJob.
* Use entrypoint instead moudle_name and class_name
* Support Type hint for Builder

Better alignment with internal changes

Unified Builder Api Changes.

ut.


dlrover/python/unified/api/builder/rl.py
```diff
@@ -10,14 +10,35 @@
...
     DLStreamType,
     DLType,
     RLRoleType,
 )
 
-from .base import DLJobBuilder
+from .base import DLJob, DLJobBuilder
+
+
+class RLJob(DLJob):
```
BalaBalaYi, line 24: add a example using api(in tutorial and examples)

PR #1620, opened 2025-08-26
remove redundancy save_model()

qijiale76:

Remove the redundant `save_model()` function call. We only use [`self.flash_checkpointer.save_checkpoint_to_storage()`](https://github.com/intelligent-machine-learning/dlrover/blob/857e5caa5bb26d624c8dcbfc2494721d89e76910/dlrover/trainer/torch/flash_checkpoint/hf_trainer.py#L242).

The `save_model()` function was performing a synchronous write of the checkpoint to `.bin` files. This operation blocked the asynchronous workflow of Flash Checkpoint. Removing this redundant call improves the efficiency.

No.

Test env: dlrover + transformers + DeepSpeed


BalaBalaYi:
My current personal opinion:

1. The flash ckpt implementation is tightly coupled to existing open-source implementations.
2. This part of the implementation has not been maintained for a long time.
3. The reason the user raised the current issue is that, based on my recent investigation, the source code it originally depended on has changed significantly.
4. Therefore, I do not recommend users rely on this implementation; if it must be used, please conduct more extensive compatibility testing.
5. I have no objections to the current changes, but it would be best to specify the dependency versions of the test environment.

PR #1626, opened 2025-09-01
【UNIFIED】Support ray node scaler(basic definition + default impl)

BalaBalaYi:

1. Node replacement might need some node info from actor, so add 'NodeInfo' and 'get_node_info' implementations.
2. Add 'relaunch_node_if_needed' in controller's manager for node relaunching during restarting.
3. Add limit on node relaunch.

To support the node replacement capability when using Ray, a corresponding RayNodeScaler and the relevant logic have been added. 

However, considering that Ray's node replacement is highly dependent on Ray's deployment, no default implementation is provided. (Internally the node replacement is implemented by integrating with KubeRay)

No

UT


dlrover/python/unified/common/constant.py
```diff
@@ -13,6 +13,11 @@
 
 
 RAY_HANG_CHECK_INTERVAL = 5.0  # Interval for monitoring actor invocations
+RAY_SINGLE_NODE_RELAUNCH_WAIT_TIME = 60
+
+
+class MasterConstant(object):
+    MASTER_STATE_KEY_PREFIX = "runtime_context_state_"
```
way-zer: add dlrover prefix to value?

dlrover/python/unified/controller/auto_registry.py
```diff
@@ -0,0 +1,161 @@
...
+import importlib
+import importlib.util
+import inspect
+import pkgutil
+from typing import Dict, Type, List, Set, Optional
+
+from dlrover.python.common.log import default_logger as logger
+
+
+class AutoExtensionRegistry:
```
way-zer: Move to util?

dlrover/python/unified/controller/schedule/graph.py
```diff
@@ -55,6 +55,18 @@ def spec(self) -> WorkloadDesc:
     def to_actor_info(self) -> "ActorInfo":
         """Convert to NodeInfo. Exposed to workers and sub-masters."""
 
+    def inc_restart_count(self):
```
way-zer, line 82: I prefer Vertex as a data class, no bussiness function, directly update by manager.

dlrover/python/unified/common/actor_base.py
```diff
@@ -141,6 +161,20 @@ def report_actor_restarted(self, name: str):
         # default delegate to master
         PrimeMasterApi.report_actor_restarted(name)
 
+    def get_node_info(self):
+        """Get the current actor's ray node's information."""
+
+        node_id = "unknown"
+        if ray.is_initialized() and ray.get_runtime_context():
```
way-zer: Inside ActorBase, ray is always `initialized`. No need to check.

dlrover/python/unified/controller/auto_registry.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    Auto-discovery extension registry. It automatically discovers abstract
+    classes decorated with @extensible and implementation classes decorated
```
way-zer: Try not use reflect?

Create a `Extensible` base class, like:
```python
class Extensible:
    _extensions: ClassVar[list[type]]

    @classmethod
    def register_extension(cls, ext: type, priority: int = 0) -> None: ...

    @classmethod
    def get_extensions(cls) -> list[type]: ...
```

dlrover/python/unified/common/actor_base.py
```diff
@@ -46,6 +47,25 @@ class ActorInfo:
...
+
+    id: str
+    hostname: Optional[str] = None
+    ip_address: Optional[str] = None
+    envs: Optional[dict[str, str]] = None
+
+    def __str__(self):
+        return f"NodeInfo(id='{self.id}', hostname={self.hostname}, ip_address='{self.ip_address}')"
+
+    def __hash__(self):
```
way-zer, line 62: remove `__hash__` and `__eq__`.

dlrover/python/unified/controller/manager.py
```diff
@@ -12,58 +12,106 @@
...
         self.scheduler: Scheduler = Scheduler(config)
-        self.sync = SyncManager()
+        self.scaler: BaseRayNodeScaler = get_scaler(config)
+        self.sync_manager = SyncManager()
 
         # Runtime state
-        self._stage: MasterStage = MasterStage.INIT
+        is_init, self._context = self._init_or_reload()
+        if is_init:
+            self.save()
```
way-zer, line 91: No need save, as another `init` will get same result.

dlrover/python/unified/common/constant.py
BalaBalaYi: done

dlrover/python/unified/common/actor_base.py
BalaBalaYi: done

dlrover/python/unified/controller/auto_registry.py
BalaBalaYi: done

dlrover/python/unified/common/actor_base.py
BalaBalaYi: used for set operations

dlrover/python/unified/controller/manager.py
BalaBalaYi: 'as another init will get same result', it's because the current other runtime parts have not implemented save yet

dlrover/python/unified/controller/manager.py
```diff
@@ -153,27 +201,117 @@ async def _main_loop(self):
...
-            if actor.restart_count > actor.spec.max_restart:
-                self.request_stop(
-                    f"Actor {actor.name} has exceeded the maximum restart count: {actor.restart_count}."
-                )
-                return
+            actor.inc_restart_count()
+
+        # relaunch node 1st if needed
+        try:
+            await self.relaunch_node_if_needed(actors)
```
way-zer: if request_stop inside, you also need return here.

dlrover/python/unified/controller/manager.py
```diff
@@ -219,12 +357,36 @@ async def wait(self):
...
-            f"nodes: {[node.name for node in self.graph.vertices]}"
-        )
-        # TODO implement actual save logic. (Ref state_backend.py)
+
+        state_key = self._get_state_key()
+        self._state_backend.set(state_key, self._context.serialize())
+
+        logger.info("Save runtime context into state.")
+
+    def _init_or_reload(self) -> Tuple[bool, ManagerRuntimeContext]:
```
way-zer, line 434: May be it's better keep this function simpler, only `try_load`, keep init in `__init__`.

dlrover/python/unified/controller/auto_registry.py
BalaBalaYi: Here is an explanation of the differences between the two approaches:

**1. The Current Method**
*   **Advantage:** Extension implementations only need to be declared, making the development process simpler.
*   **Disadvantage:** During the initial load, the system needs to perform a full scan to find all matches. This results in an additional latency of a few seconds.

**2. The Method You Mentioned**
*   **DisAdvantage:** The implementation of extensions becomes more complex. Besides the declaration, manual registration is required.
*   **Advantage:** Because registration is done manually, the system naturally avoids the need for an additional polling/scanning process, eliminating the associated performance cost.

dlrover/python/unified/controller/manager.py
BalaBalaYi: need to ````try_load``` 1st , than ````try_init``` if so, still need 2 conditions check to ensure 'init' will not override the 'reload' state

dlrover/python/unified/controller/schedule/graph.py
```diff
@@ -188,3 +218,26 @@ def create(cls, dl_config: DLConfig) -> "DLExecutionGraph":
...
+        )
+
+    def get_all_actors_by_specified_nodes(self, nodes: set[str]):
+        return [
+            actor
+            for actor in self.vertices
+            if actor.node_info is not None and actor.node_info.id in nodes
+        ]
+
+    def update_actor_failure(self, name):
```
way-zer: avoid use name, always use Vertex inside manager.

dlrover/python/unified/util/auto_registry.py
```diff
@@ -0,0 +1,163 @@
...
+        """
+        Do the auto-discovery for the extension package.
+
+        Args:
+            extension_package: extension package name
+        """
+
+        try:
+            # scan basics and do registering
+            basics = cls._scan_package_for_extensions("dlrover.python")
```
way-zer, line 46: Maybe import many unnesserary module?

dlrover/python/unified/backend/elastic/master.py
```diff
@@ -25,7 +25,7 @@ def _setup(self):
 
     # Lifecycle Hooks
 
-    def status(self):
+    def stage(self):
```
way-zer: rename

dlrover/python/unified/controller/manager.py
way-zer: failover only support in RUNNING stage, get a INIT save is useless.

dlrover/python/unified/controller/manager.py
```diff
@@ -74,14 +131,16 @@ async def prepare(self):
...
 
         # Wait for all nodes to be ready
-        await self._wait_ready([node.name for node in self.graph.vertices])
+        actors = [actor.name for actor in self.graph.vertices]
+        await self._wait_ready(actors)
         self._update_stage(MasterStage.READY)
+
         await self._nodes_check()
 
     async def _wait_ready(self, actors: List[str]):
```
way-zer: As this function become complex, it's better passing List[Vertex]

dlrover/python/unified/controller/manager.py
```diff
@@ -95,6 +154,13 @@ async def _wait_ready(self, actors: List[str]):
             )
             await asyncio.sleep(5)
 
+        # update all actor's node info
+        node_info_res = await invoke_actors_t(
+            remote_call.get_node_info, actors
+        )
+        for actor_name, node_info in node_info_res.as_dict().items():
+            self.graph.by_name[actor_name].set_node(node_info)
```
way-zer: Avoid use by_name inside manager

dlrover/python/unified/controller/manager.py
way-zer: try_load return Optional, then `try_load() or ManagerRuntimeContext()` directly

dlrover/python/unified/controller/manager.py
```diff
@@ -153,28 +219,174 @@ async def _main_loop(self):
...
+        reasons should be filtered out by the following `actor.restarting`
+        logic.
+        """
+
+        if actor.restarting:
+            return  # Actor is already restarting, no need to handle it again.
+
+        name = actor.name
+        # record restarted(failure) actor
+        self.graph.update_actor_failure(name)
```
way-zer: not use name, directly use vertex

dlrover/python/unified/controller/manager.py
```diff
@@ -153,28 +219,174 @@ async def _main_loop(self):
...
+                f"total relaunching: {len(relaunch_nodes)}, "
+                f"finish relaunching: {current_relaunched_num}."
+            )
+            await asyncio.sleep(wait_interval)
+            current_relaunched_num = get_relaunched_num()
+
+        self._update_node_restart_count(len(relaunch_nodes))
+        logger.info(f"All target nodes relaunched: {relaunch_nodes}")
+
     async def restart_actors(self, actor_names: List[str]) -> None:
```
way-zer: pass actors instead actor_names, as the function become complex.

dlrover/python/unified/controller/auto_registry.py
way-zer: The Current Method is much bader, as it depends on module path/reflect, and it's annotation style. 
Use decorator is better, it could register, no need manual registration.

way-zer:
Another thing need discuss: failure should be associated to Node instead Actor?

way-zer (approved):

PR #1628, opened 2025-09-03
Skip container with empty cpu or mem when calculate node resources from yaml

majieyue:

Skip container with empty cpu or mem when calculate node resources from yaml

In some cases, the container may have empty cpu or mem in requested resources

No

UT


dlrover/python/master/watcher/k8s_watcher.py
```diff
@@ -134,7 +134,12 @@ def _convert_pod_yaml_to_node(pod):
     pod_id = int(metadata.labels[replica_index_key])
     rank_id = int(metadata.labels[rank_index_key])
     relaunch_count = int(metadata.labels[relaunch_count_key])
-    resource = _parse_container_resource(pod.spec.containers[0])
+    resource = NodeResource(0, 0)
+    for container in pod.spec.containers:
+        res = _parse_container_resource(container)
+        if res.cpu > 0 and res.memory > 0:
+            resource = res
```
BalaBalaYi, line 146: It is not appropriate to directly judge if the value is >0. 

Since a node is equivalent to a pod and this resource field is used to create a pod, I think it is acceptable to directly take the maximum value. Additionally, the GPU attribute should be included.

dlrover/python/master/watcher/k8s_watcher.py
```diff
@@ -134,7 +134,12 @@ def _convert_pod_yaml_to_node(pod):
     pod_id = int(metadata.labels[replica_index_key])
     rank_id = int(metadata.labels[rank_index_key])
     relaunch_count = int(metadata.labels[relaunch_count_key])
-    resource = _parse_container_resource(pod.spec.containers[0])
+    resource = NodeResource(0, 0)
+    for container in pod.spec.containers:
```
BalaBalaYi, line 143:
```suggestion
     # this is a temporary workaround to retrieve the 'main' container's resource
    for container in pod.spec.containers:
```

dlrover/python/common/node.py
```diff
@@ -189,7 +192,7 @@ def __init__(
         rank_index=None,
         relaunch_count=0,
         critical=False,
-        max_relaunch_count=0,
+        max_relaunch_count=_dlrover_context.max_relaunch_count,
```
way-zer, line 195: Maybe we should distinguish between NodeState and the Node object used for transmission.

And it is strange that the default value of the data object is from context. 
Also need notice that the default value will be fixed when class defining.

dlrover/python/common/global_context.py
```diff
@@ -143,6 +144,7 @@ def __init__(self):
         self.npu_per_node = DefaultValues.NPU_NUM_PER_NODE
         # pre-check args
         self.pre_check_operators = DefaultValues.PRE_CHECK_OPS
+        self.max_relaunch_count = DefaultValues.MAX_RELAUNCH_COUNT
```
BalaBalaYi, line 147: u need to modify this part: https://github.com/intelligent-machine-learning/dlrover/blob/30ae58ca998d2bc0e598254a5a2c3df6bc7574ed/dlrover/python/master/node/dist_job_manager.py#L166

dlrover/python/master/node/dist_job_manager.py
```diff
@@ -602,6 +600,8 @@ def _process_list_nodes(self, nodes: List[Node]):
                     )
                     new_node = copy.deepcopy(node)
                     self._job_context.update_job_node(new_node)
+                    if new_node.has_group():
+                        self._job_context.update_job_node_by_group(new_node)
```
way-zer, line 604: merge to `update_job_node`?

All `update_job_node_by_group` usages are in same pattern:

```python
self._job_context.update_job_node(new_node)
if new_node.has_group():
    self._job_context.update_job_node_by_group(new_node)
```

BalaBalaYi, line 604: the new_node will be deep copy again in method 'update_job_node_by_group', should rm the outside 'deep.copy'

suggestion:
```
self._job_context.update_job_node(node)
if node.has_group():
    self._job_context.update_job_node_by_group(node)
```

majieyue: leave it as TODO action, which will combine update_job_node_by_group with update_job_node

majieyue: leave it as TODO action, which will combine update_job_node_by_group with update_job_node

PR #1634, opened 2025-09-09
[Unified] Add events

way-zer:

Please describe the changes you have made or proposed in this pull request.

Specify whether this pull request introduces any changes that users will directly interact with or notice.

Detail the testing process you have undertaken to ensure the changes in this pull request are valid and working as intended.

### DIFF
--- a/dlrover/python/unified/util/async_helper.py
+++ b/dlrover/python/unified/util/async_helper.py
@@ -13,6 +13,7 @@
 
 import asyncio
 from concurrent.futures import Future
+from threading import Thread
 from typing import Any, Coroutine, TypeVar
 
 """Utility functions for asynchronous operations in DLRover.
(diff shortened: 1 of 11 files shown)


dlrover/python/unified/util/async_helper.py
```diff
@@ -37,6 +38,42 @@ def init_main_loop():
         )
 
 
+def unsafe_run_blocking(coro: Coroutine[Any, Any, T]) -> T:
+    """Run a coroutine in a separate thread and wait for its result.
+
+    Note: All coroutines should be in the main loop, unless could not use main loop.
```
BalaBalaYi:
```suggestion
    Note: All coroutines should be in the main loop of the target thread, unless could not use main loop(of the target thread).
```

PR #1643, opened 2025-09-24
Avoiding socket file conflicts

zhengchenyu:

Add node rank as part of the socket file name.

When I run two dlrover app to simulate multi host environment on one machine, it causes socket file conflicts.

No.

Real job.

### DIFF
--- a/dlrover/python/common/multi_process.py
+++ b/dlrover/python/common/multi_process.py
@@ -26,6 +26,7 @@ from typing import Dict
 
 import _posixshmem
 
+from . import env_utils
 from .constants import NodeEnv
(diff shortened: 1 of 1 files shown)


BalaBalaYi (changes requested):
First, the job name must be unique. If it is not, problems will arise not only here. 
Second, if the job name is unique, this issue simply wouldn't exist here. (The lifecycle of a job is completely bound to the pod.)

zhengchenyu:
@BalaBalaYi
I'm not saying this will happen with real jobs. It's just that there will be conflicts when debugging on a single machine. I added this just to make it work in all modes.

PR #1644, opened 2025-09-24
Support failover for job failure.

BalaBalaYi:

1. Add 'failover_trigger_strategy' and 'failover_exec_strategy' in job config. 
3. Trigger failover in controller(manager) 's main loop.

Support job failover when job has failed result.

No

UT

### DIFF
--- a/dlrover/python/unified/controller/manager.py
+++ b/dlrover/python/unified/controller/manager.py
@@ -456,7 +456,7 @@ class PrimeManager:
             # Don't support failover in other stages, which are little possibility
             logger.warning(f"Job is in stage {self.stage}, terminating.")
             ControllerEvents.failover_stop(self.stage)
-            self._clear_failover_stage(InternalDLWorkloadRole.GLOBAL_ROLE)
+            self._clear_failover_stage()
             self._do_stop()
 
     async def deal_with_actor_finished(


dlrover/python/unified/controller/manager.py
```diff
@@ -186,9 +186,26 @@ async def _main_loop(self):
             # all driver roles finished
             if all(result is not None for result in results):
                 if any(result == ExecutionResult.FAIL for result in results):
-                    self.request_stop(
-                        "All driver roles finished, but some nodes failed."
-                    )
+                    if self.config.failover_strategy_when_failed == 1:
```
way-zer: It's better handle in `deal_with_actor_finished` before `actor.result` update, as it's common that one worker fails (other may hang) and this branch need all roles finished.

BalaBalaYi: > It's better handle in `deal_with_actor_finished` before `actor.result` update, as it's common that one worker fails (other may hang) and this branch need all roles finished.

It doesn't seem appropriate to place it in ```deal_with_actor_finished```. This is because the current minimum granularity for failover should be at the role level, followed by the job level. Therefore, it seems more appropriate to place it within the operational logic flow for role-level results rather than in ```deal_with_actor_finished```, which is at the worker level.

BalaBalaYi: @way-zer any questions?

way-zer: Handle in `deal_with_actor_finished`, as it's the event handler for `fail`.

This does not conflict with role-level failover. Detect failure at worker-level and then resolve the failure(failover) at the role-level.

BalaBalaYi: > Handle in `deal_with_actor_finished`, as it's the event handler for `fail`.
> 
> This does not conflict with role-level failover. Detect failure at worker-level and then resolve the failure(failover) at the role-level.

The current implementation is exactly as you described.: 
1. Detect failure at the worker-level in `deal_with_actor_finished`: performs setting the worker-level result, which I believe is sufficient.
2. Then resolve the failure (failover) at the role-level in the main loop. 

It is important to note that `deal_with_actor_finished` should complete the update of the actor's finished result and should not be coupled with failover operations. The specific logic for failover operations is indirectly related to the results updated by `deal_with_actor_finished`.

way-zer: I understand your intention: you want to handle failover based on state rather than in the `deal_with_actor_finished` event.

However, there are some issues with this approach.
1. It may be difficult to achieve your desired state (all workers finished). A single worker may fail while other workers/roles hang.
2. The root cause of the failure, namely the specific worker and role, is lost in the main loop. If all workers have finished, a role-level restart is meaningless.

way-zer: My idea:
When any worker fails, immediately initiate failover according to the policy, performing a role/job-level restart. This restart would directly disrupt other workers involved.

BalaBalaYi: 'It may be difficult to achieve your desired state (all workers finished). A single worker may fail while other workers/roles hang.' 

This is indeed a fairly clear rationale. The policy of 'When any worker fails, immediately initiate failover according to the policy'—whether to directly trigger fault tolerance upon any error—should be reconsidered to expose a job-level option. If the option is set to true, then trigger immediately; otherwise, wait to trigger when handling the result in the main loop.

way-zer: It is reasonable to add an option. 

Perhaps this can still be implemented in the main loop. The pseudocode is as follows:
```
foreach role:
  if role.anyWorkerFaied:
    if option.immedatieFailover or role.allWorkerFinished:
      Do Job/Role-level Failover.
```

BalaBalaYi: @way-zer updated

dlrover/python/unified/controller/manager.py
```diff
@@ -186,8 +196,8 @@ async def _main_loop(self):
             # all driver roles finished
             if all(result is not None for result in results):
                 if any(result == ExecutionResult.FAIL for result in results):
-                    self.request_stop(
-                        "All driver roles finished, but some nodes failed."
+                    await self._process_failover(
```
way-zer, line 208: Seems not match arguments
```
async def _process_failover(self, role_name="", reason="")
```

dlrover/python/unified/controller/manager.py
```diff
@@ -177,6 +177,16 @@ async def _main_loop(self):
         while self.stage == MasterStage.RUNNING:
             await self._notify_main_loop.acquire()
 
+            any_failure = [
+                role.has_any_failure() for role in self.graph.roles.values()
+            ]
+            if any(any_failure):
```
way-zer, line 192: Need to check “Is it safe to enter multiple times?”, As it's in loop

BalaBalaYi: > Need to check “Is it safe to enter multiple times?”, As it's in loop

Prefer to use a status flag instead, placed in `_process_failover`, but the implementation of this flag may also need to be compatible with role-level failover scenarios.

way-zer: Yeah, a flag looks good. 

In addition, I am worried that if other conditions are added in the future, `logger.info` will be too frequent. 
Maybe it's better to add `asyncio.sleep(1)` after loop to achieve throttling.

dlrover/python/unified/controller/manager.py
BalaBalaYi: fixed

dlrover/python/unified/controller/manager.py
BalaBalaYi: updated

dlrover/python/unified/controller/manager.py
```diff
@@ -445,24 +455,55 @@ async def deal_with_actor_finished(
...
+        self, role_name="global", reason="unknown reason"
+    ):
+        if self.is_failover_stage(role_name):
             logger.info(
-                f"Role level failover is not supported yet, do job failover instead, reason: {reason}."
+                "Failover is already in progress, skip the following process."
             )
-            await self.restart_job()
+            return
         else:
```
way-zer, line 504: `else` is not necessary for Short-circuit if

dlrover/python/unified/controller/manager.py
```diff
@@ -50,6 +51,13 @@ class RuntimeState:
     """
 
     stage: MasterStage = MasterStage.INIT
+    # lag for failover status, the presence of a corresponding key-value pair
+    # indicates that failover is in progress.
+    # format:
+    # key: role name, is empty if job level
```
way-zer: It's `global` for job level in code?

dlrover/python/unified/controller/manager.py
```diff
@@ -420,12 +441,22 @@ def self_recover(self):
             self._task = asyncio.create_task(
                 self._main_loop(), name="job_monitor"
             )
+
+            # if a master failover and a failover of any role overlap,
+            # we currently perform a global restart again for safety.
```
BalaBalaYi, line 446: @way-zer A safety measure added, please review.

PR #1645, opened 2025-09-24
Configure the timeout and retry count between the master and worker.

zhengchenyu:

Configure the below parameter which is constant before.

* the timeout of worker connect to master.
* the retry time of worker connect to master.
* the timeout of the master to check the node alive.

In some scenarios, these non-configurable default values ​​are set too large, resulting in, for example, a 10-minute hang time.

Add all config have default value. 

Real job.

### DIFF
--- a/dlrover/python/elastic_agent/torch/training.py
+++ b/dlrover/python/elastic_agent/torch/training.py
@@ -218,6 +218,8 @@ class ElasticLaunchConfig(LaunchConfig):
     training_log_file: str = ""
     failure_node_errors: str = ""
     numa_affinity: bool = False
+    connect_master_timeout = 300
+    connect_master_max_retry = 5
 
     def set_node_unit(self, node_unit):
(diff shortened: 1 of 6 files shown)


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -218,6 +218,8 @@ class ElasticLaunchConfig(LaunchConfig):
     training_log_file: str = ""
     failure_node_errors: str = ""
     numa_affinity: bool = False
+    connect_master_timeout = 300
```
BalaBalaYi, line 221: It's unnecessary to define these two parameters here. You can refer to the timeout mechanism inside the master server/client and leverage existing parameters (by exposing them via config). 

Additionally, note that these parameters should not be passed through dlrover-run; instead, they should be directly obtained by the master through job args.

BalaBalaYi (changes requested):

PR #1649, opened 2025-09-26
avoid save_to_memory running too fast then causing save_to_storage to fail

qijiale76:

This PR based on #1585

Previously, a save_to_memory call could overwrite a shared memory buffer while the background saver process was still reading from it for a save_to_storage operation. To prevent this, a notification queue was introduced in #1585, where the saver process notifies the training process once the buffer is free.

However, the implementation in #1585 only had rank 0 consume this notification. This caused all non-rank-0 processes to hang indefinitely on subsequent checkpoints, as they never consumed the event from their corresponding notification queue.

This patch fixes the issue by ensuring that every rank waits for and consumes the notification from its dedicated queue before proceeding with the next save_to_memory operation. This guarantees that the shared memory buffer is not written to prematurely.

Without proper synchronization across all ranks, tasks would hang during checkpointing, especially when checkpoints were triggered in rapid succession. This fix ensures the reliability of the asynchronous saving process by correctly coordinating the main training processes and the background saver.

No

Verified in live training jobs. The race condition is resolved and the task no longer hangs.

### DIFF
--- a/dlrover/python/elastic_agent/torch/ckpt_saver.py
+++ b/dlrover/python/elastic_agent/torch/ckpt_saver.py
@@ -86,6 +86,11 @@ class CheckpointEvent:
 
 
 @dataclass
+class CheckpointNotifyEvent:
+    step: int = 0
+
+
+@dataclass
 class TensorMeta:
     shape: Tuple[int] = None  # type: ignore
     dtype: torch.dtype = None  # type: ignore
@@ -451,7 +456,11 @@ class AsyncCheckpointSaver(metaclass=ABCMeta):
         self._latest_step = 0
         qname = CheckpointSharedObjPrefix.SAVE_STEP_QNAME + str(0)
         self._event_queue = SharedQueue(name=qname, create=True)
+        self._notify_queues = []
         for i in range(self.local_shard_num):
+            self._notify_queues.append(
+                SharedQueue(name=CheckpointSharedObjPrefix.SAVE_STEP_QNAME + "_notify_" + str(i),
+                            create=True))
             self._shm_handlers.append(SharedMemoryHandler(i))
             lock_name = CheckpointSharedObjPrefix.SHM_LOCK_NAME + str(i)
             self._shm_locks.append(SharedLock(name=lock_name, create=True))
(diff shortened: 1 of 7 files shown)


dlrover/python/elastic_agent/torch/ckpt_saver.py
```diff
@@ -451,7 +456,11 @@ def __init__(
         self._latest_step = 0
         qname = CheckpointSharedObjPrefix.SAVE_STEP_QNAME + str(0)
         self._event_queue = SharedQueue(name=qname, create=True)
+        self._notify_queues = []
         for i in range(self.local_shard_num):
+            self._notify_queues.append(
```
BalaBalaYi, line 464: use a function to generate queue name

dlrover/trainer/torch/flash_checkpoint/fsdp_engine.py
```diff
@@ -469,7 +469,7 @@ def get_saving_ranks(self):
         return None
 
     @timer
-    def save_to_memory(self, step, state_dict, paths: Dict[str, str]):
+    def _save_to_memory(self, step, state_dict, paths: Dict[str, str]):
```
BalaBalaYi, line 472: This should not be modified, as it is inherently a public abstract method.

dlrover/trainer/torch/flash_checkpoint/deepspeed_engine.py
```diff
@@ -89,6 +89,10 @@ def save_to_memory(self, step, state_dict, paths):
                 ["model_states", "optim_states"] of the state dict and
                 the value is the path of storage to save.
         """
+        if self._checkpoint_event_step > 0:
```
BalaBalaYi, line 93: Should be basic impl from parent class.

dlrover/python/elastic_agent/torch/ckpt_saver.py
```diff
@@ -85,6 +85,11 @@ class CheckpointEvent:
     global_shard_num: int = 0
 
 
+@dataclass
+class CheckpointNotifyEvent:
```
BalaBalaYi, line 89: This is not used?

dlrover/trainer/torch/flash_checkpoint/deepspeed_engine.py
BalaBalaYi, line 93: Also i don't get the point why not using 'last_step' but use '_checkpoint_event_step' instead?

BalaBalaYi (changes requested):

PR #1653, opened 2025-10-18
Add dlrover_run_affinity.sh before the entrypoint, to run each process with its numa affinity based on its LOCAL_RANK

majieyue:

The running processes is designated one for each GPU, the its numa affinity is emtpy.

Add numa affinity for each running process

No

UT

### DIFF
--- a/dlrover/python/master/node/worker.py
+++ b/dlrover/python/master/node/worker.py
@@ -668,10 +668,28 @@ class WorkerManager(TrainingNodeManager):
         nodes = self._get_nodes()
         return all([node.is_node_check_failed() for _, node in nodes.items()])
 
-    def is_all_initial_workers_node_check_failed(self, worker_num: int):
+    def is_all_initial_workers_node_check_failed(
+        self, worker_num: int, min_worker_num: int = 4
+    ):
         """
-        Check all initial workers are check-failed
-        (exclude new relaunched workers)"""
+        Check all initial workers are check-failed(exclude new relaunched workers).
+
+        Will skip judgement if worker number is less than min_worker_num because
+        when the number of workers is too small (especially in single-machine
+        or dual-machine scenarios), the probability of an all check failure
+        increases significantly. At the same time, in scenarios with
+        fewer workers, the number of fault tolerance attempts is limited by
+        the total number of workers, which does not lead to excessively high
+        fault tolerance costs. Hence, this boundary restriction is imposed.
+        """
+
+        if worker_num < min_worker_num:
+            logger.debug(
+                "Skip all initial workers node-check failed judgement "
+                f"for worker num: {worker_num} < {min_worker_num}."
+            )
+            return False
+
         nodes = [
             node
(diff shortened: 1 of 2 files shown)


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -1015,14 +1018,18 @@ def _invoke_run(self, role: str = DEFAULT_ROLE) -> RunResult:
             f"{spec.get_entrypoint_name()}"
         )
 
-        self._initialize_workers(self._worker_group)
+        if self._config.numa_affinity and isinstance(spec.entrypoint, str):
+            wg = self._worker_group
+            logger.info(f"WorkerGroup before numa affinity: {wg.spec} {wg}")
```
BalaBalaYi: set as debug?

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -536,6 +537,7 @@ def __init__(
         self._node_rank = node_rank
         self._config = config
         self._entrypoint = entrypoint
+        self._args = spec
```
BalaBalaYi: not used?

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -1015,14 +1018,18 @@ def _invoke_run(self, role: str = DEFAULT_ROLE) -> RunResult:
...
         )
 
-        self._initialize_workers(self._worker_group)
+        if self._config.numa_affinity and isinstance(spec.entrypoint, str):
+            wg = self._worker_group
+            logger.info(f"WorkerGroup before numa affinity: {wg.spec} {wg}")
+            wg.spec.args = (wg.spec.entrypoint,) + wg.spec.args
+            wg.spec.entrypoint = ScriptPath.RUN_AFFINITY_SCRIPT
+            logger.info(f"WorkerGroup after numa affinity: {wg.spec} {wg}")
+            self._initialize_workers(wg)
```
BalaBalaYi: Why distinguish here? Why not just directly call `self._initialize_workers(self._worker_group)`?

majieyue: because self._worker_group will affect not only elastic agent, but node check agent as well. we don't want to change node check behavior

dlrover/python/elastic_agent/torch/training.py
majieyue: INFO looks appropriate. The logs only print once

dlrover/python/elastic_agent/torch/training.py
majieyue: fixed

dlrover/python/elastic_agent/torch/training.py
BalaBalaYi: I mean, the worker_group is an object. Whether you retrieve it to modify or directly modify self._worker_group, you're operating on the same object.

PR #1657, opened 2025-10-30
Optimize check failed rejection.

BalaBalaYi:

Add a worker-num judgement on method 'is_all_initial_workers_node_check_failed'.

Will not intercept the job running if all node-check failed in few workers scenarios.

No

UT


way-zer (approved):
LGTM

PR #1665, opened 2025-11-12
Fix node failure reason reporting.

BalaBalaYi:

Override node exit reason by self detected reason.

Fix: detected exit reason not reported 

No

Training test with node check failed case.


majieyue (approved):
lgtm

PR #1674, opened 2025-12-15
Support k8s util extension in watcher.

BalaBalaYi:

1. Add a overridable 'BaseK8sUtil' for util extension used in k8s watcher.
2. Clean up internal constant using(including 'sigma').
3. Add a extension point for getting specified exit reason from pod info.

To enhance getting pod's exit reason.

No

UT and Training with specified pod annotations.

### DIFF
--- a/dlrover/python/common/global_context.py
+++ b/dlrover/python/common/global_context.py
@@ -281,7 +281,7 @@ class Context(Singleton):
                 break
         return False
 
-    def get_k8s_util_impl(self):
+    def get_k8s_util(self):
         """Can be overridden by subclasses."""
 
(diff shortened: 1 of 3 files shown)


dlrover/python/common/global_context.py
```diff
@@ -279,3 +280,8 @@ def is_pre_check_operator_bypass(self, pre_check_op) -> bool:
                         return True
                 break
         return False
+
+    def get_k8s_util_impl(self):
```
samplise: The name is a little wired. `k8s_util` is a more common use case.

samplise (approved):

PR #1682, opened 2026-01-09
Enhancement failover processing with node relaunching.

BalaBalaYi:

1. 'process_failover' from 'worker fail result' will do node relaunch(if needed) before 'restart_job'.
2. if actor restarting is caused by node-relaunch, should not invoke 'do_failover'
3. split node relaunch impl into 1 base function: '_relaunch_nodes' and 2 usage function: '_relaunch_single_node' and '_relaunch_node_group'
4. support node-group failover when group label is configured 

1. Fix: there is no space for node relaunch when worker result is FAIL.
6. Support: node-group failover procedure.
7. Other enhancement.

please refer to: [node_group_using.md](https://github.com/intelligent-machine-learning/dlrover/pull/1682/files#diff-0e39165f4135f86c342c77d3dac4845c13372b65ec5c0a1e48364ef54f9e7eb6)

UT and Training Job with different type failure.

### DIFF
--- a/dlrover/python/unified/controller/manager.py
+++ b/dlrover/python/unified/controller/manager.py
@@ -542,10 +542,6 @@ class PrimeManager:
                 await self._task
             except asyncio.CancelledError:
                 logger.info("Monitor task cancelled, proceeding with restart.")
-            except RuntimeError:
-                logger.warning(
-                    f"Unexpected runtime error when await task: {type(self._task)}"
-                )
 
             logger.info("Restarting all actors...")
             await self.restart_actors(self.graph.vertices)


dlrover/python/unified/controller/manager.py
```diff
@@ -92,6 +95,8 @@ def __init__(self, config: JobConfig) -> None:
         self.state: RuntimeState = self._load_state() or RuntimeState(
             graph=DLExecutionGraph.create(config.dl_config),
         )
+        self._state_lock = threading.Lock()
+        self._failover_lock = threading.Lock()
```
way-zer: use `asyncio.Lock` if possible, manager is single-thread/asyncio

dlrover/python/unified/controller/manager.py
```diff
@@ -387,13 +496,18 @@ def request_stop(self, reason: str, code: int = 1):
...
             self._update_stage(MasterStage.STOPPING)
             self._notify_main_loop.release()
         else:
             # main loop is not running, do stop directly
             self._do_stop()
 
+    def _in_running_stage(self):
+        """Including running and failover stages."""
+
+        return self.stage in (MasterStage.RUNNING, MasterStage.FAILOVER)
```
way-zer: Do FAILOVER must be one MasterStage?

How about Running with another bool flag?

dlrover/python/unified/common/workload_desc.py
```diff
@@ -173,6 +173,14 @@ class BaseWorkloadDesc(BaseModel, ABC):
             "If False, job will not wait, suitable for RPC driven workloads."
         ),
     )
+    node_group_failover: Tuple[bool, str, int] = Field(
```
way-zer: Create a new Model with label and timeout?
And `None` represet disabled?

dlrover/python/unified/controller/manager.py
BalaBalaYi: Indeed, this part may need some discussion. The main reason for introducing this new state currently is to address race conditions in the fault tolerance process. After the changes, we indeed found that the modifications have a relatively significant impact, and the main loop also needs to be adapted. Additionally, it seems feasible to leverage the previous failover_stage for judgment.

dlrover/python/unified/controller/manager.py
BalaBalaYi: these 2  locks are all used in sync function

dlrover/python/unified/controller/manager.py
BalaBalaYi: please review again

dlrover/python/unified/common/workload_desc.py
BalaBalaYi: done

dlrover/python/unified/backend/common/base_worker.py
```diff
@@ -145,9 +145,9 @@ def _execute_user_function(self, user_func):
             logger.exception(
                 f"Unexpected error occurred while executing user function({user_func})."
             )
-            self._on_execution_end(ExecutionResult.FAIL)
+            self._on_execution_end(ExecutionResult(ExecutionResultType.FAIL))
         else:
-            self._on_execution_end(ExecutionResult.SUCCESS)
+            self._on_execution_end(ExecutionResult())
```
way-zer: Explicitly "success" makes it clearer to read.

dlrover/python/unified/common/actor_base.py
```diff
@@ -64,6 +69,48 @@ class NodeInfo:
     ip_address: Optional[str] = None
     envs: dict[str, str] = field(default_factory=dict)
 
+    def __repr__(self) -> str:
+        return f"NodeInfo(id='{self.id}', hostname={self.hostname}, ip_address={self.ip_address})"
+
+
+@dataclass(frozen=True)
+class ExecutionResult:
```
way-zer, line 91: It appears that the Actor itself doesn't use this struct (only ExecutionResultType). Is this considered an internal struct of the manager, and therefore shouldn't the actor be modified?

dlrover/python/unified/controller/manager.py
```diff
@@ -309,52 +377,157 @@ async def _do_failover(self, actor: DLExecutionVertex):
...
-                return
-            # Ensure the nodes are removed from Ray cluster, avoid exceptions during restart processing.
-            await asyncio.wait_for(
-                wait_ray_node_remove([n.id for n in relaunched_nodes]),
-                timeout=(
-                    len(relaunched_nodes) * RAY_SINGLE_NODE_RELAUNCH_WAIT_TIME
-                ),
+            total_nodes_size = len(ray.nodes())
+
+            async def _relaunch_and_wait():
```
way-zer, line 404: Move out?

dlrover/python/unified/controller/manager.py
```diff
@@ -463,34 +652,104 @@ async def deal_with_actor_finished(
...
+            else:
+                if (
+                    InternalDLWorkloadRole.GLOBAL_ROLE
+                    in self.state.failover_stage
+                    or role_name in self.state.failover_stage
+                ):
+                    return True
+                return False
+
+    def _record_failure(self, role_name):
```
BalaBalaYi, line 687: @way-zer priority is used here

dlrover/python/unified/controller/manager.py
```diff
@@ -387,13 +566,18 @@ def request_stop(self, reason: str, code: int = 1):
...
             self._update_stage(MasterStage.STOPPING)
             self._notify_main_loop.release()
         else:
             # main loop is not running, do stop directly
             self._do_stop()
 
+    def _in_running_stage(self):
+        """Including running stages."""
+
+        return self.stage == MasterStage.RUNNING
```
way-zer, line 575: Prefer to `inline`.

dlrover/python/unified/controller/manager.py
```diff
@@ -463,34 +650,104 @@ async def deal_with_actor_finished(
...
+
         logger.info(f"Actor {actor.name} reported result {result}.")
         actor.result = result
         self._notify_main_loop.release()
 
     def _set_failover_stage(self, role_name):
-        if role_name not in self.state.failover_stage:
-            self.state.failover_stage[role_name] = int(time.time())
-            logger.debug(f"Setting failover stage: {role_name}")
+        with self._failover_lock:
```
way-zer: Is this lock necessary? There is no race in synchronized methods with asyncio.

dlrover/python/unified/controller/manager.py
BalaBalaYi: Inner is better: only used here and less variable passing

dlrover/python/unified/controller/manager.py
BalaBalaYi: u r right, '_do_failover' and '_process_failover' are all async methods, there is only possible logic race condition here

dlrover/python/unified/controller/manager.py
BalaBalaYi: keep it:
1.  more readable than direct enum comparison
2. hides implementation details (currently == MasterStage.RUNNING), allows for more complex logic in the future
3. avoids code duplication
4. easy in unit testing and mocking

dlrover/python/unified/common/actor_base.py
BalaBalaYi: answered below

dlrover/python/unified/backend/common/base_worker.py
```diff
@@ -172,19 +193,17 @@ def _execute_user_command(self, command: str):
...
-            exit_code = 0 if e.code is None or e.code == 0 else 1
-
-            if exit_code == 0:
-                result = ExecutionResult()
-            else:
-                result = ExecutionResult(ExecutionResultType.FAIL)
-            self._on_execution_end(result)
+            self._on_execution_end(
+                ExecutionResultType.SUCCESS
+            ) if e.code is None or e.code == 0 else self._on_execution_end(
```
way-zer: Format: Avoid using ternary operators for statement branches

BalaBalaYi: done

PR #1684, opened 2026-01-11
Support numa affinity bind mode

majieyue:

For numa affinity, we need to support more kinds of memory bind mode, e.g preferred, membind, and none

We need another mempolicy param to support numa memory bind mode

Yes

Job test

### DIFF
--- a/dlrover/python/tests/test_elastic_training_agent.py
+++ b/dlrover/python/tests/test_elastic_training_agent.py
@@ -868,6 +868,7 @@ class ElasticTrainingAgentRunTest(unittest.TestCase):
 
     def test_invoke_run_with_numa_affinity(self):
         self.config.numa_affinity = True
+        self.config.membind_policy = "none"
         self.config.training_port = 0
         self.spec.entrypoint = "sleep"
         self.spec.args = tuple(["3"])
--- a/dlrover/trainer/tests/torch/elastic_run_test.py
+++ b/dlrover/trainer/tests/torch/elastic_run_test.py
@@ -171,6 +171,8 @@ class ElasticRunTest(unittest.TestCase):
             "--training_port",
             "1000",
             "--numa-affinity",
+            "--membind-policy",
+            "preferred",
             "test.py",
             "--batch_size",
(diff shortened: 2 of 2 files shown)


dlrover/python/elastic_agent/torch/training.py
```diff
@@ -222,6 +222,7 @@ class ElasticLaunchConfig(LaunchConfig):
     training_log_file: str = ""
     failure_node_errors: str = ""
     numa_affinity: bool = False
+    membind_policy: str = "none"
```
BalaBalaYi, line 225: merge these 2 params into 1:
numa affinity is enabled when using 'preferred' and 'bind'

majieyue: we should also support neither prefered nor membind cases, which by default using none as no memnode binding at all. 
e.g. some hopper arch, there is no mem node and it will lead to error if set by default

BalaBalaYi: If 'membind_policy' is the only optional param, merge the current 2 params into 1:
```
...
numa_affinity_membind_policy: Optional[str] = None

@property
def numa_affinity(self): -> bool
  return self.numa_affinity_membind_policy is not None
```

Otherwise(considering there are more optional params), add specific prefix for the param(to let user know it belongings). Such as: 

```
numa_affinity_enabled = xxx
numa_affinity_membind_policy = xxx
numa_affinity_xxx = xxx
```

Also we may need a new data class for numa_affinity params later in 'ElasticLaunchConfig' if there are more optional params supported for 'numa_affinity'.

dlrover/trainer/torch/elastic_run.py
```diff
@@ -201,6 +201,15 @@ def parse_args(args):
         help="bool, set workers processes cpu numa affinity or not",
     )
 
+    parser.add_argument(
+        "--membind-policy",
+        "--membind_policy",
+        type=str,
```
BalaBalaYi, line 207: add a enum check here?

PR #1685, opened 2026-01-12
Optimize the step hang detection

majieyue:

1. reset atorch event collector first step each time rendezvous completed
2. remove max hang timeout 10min restriction

use can set up ites customized hang timeout larger than 10 min

No

UT

### DIFF
--- a/dlrover/python/elastic_agent/torch/training.py
+++ b/dlrover/python/elastic_agent/torch/training.py
@@ -222,6 +222,7 @@ class ElasticLaunchConfig(LaunchConfig):
     training_log_file: str = ""
     failure_node_errors: str = ""
     numa_affinity: bool = False
+    membind_policy: str = "none"
 
     def set_node_unit(self, node_unit):
         """Set the number unit of nodes."""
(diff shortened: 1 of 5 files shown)


dlrover/python/common/global_context.py
```diff
@@ -76,10 +76,10 @@ class DefaultValues(object):
         )
     ]
     HANG_DETECTION = 1
-    HANG_DOWNTIME = 10  # max downtime, unit is minute
+    HANG_DOWNTIME = 10  # default downtime, unit is minute
     MIN_HANG_DOWNTIME = 2  # min downtime, unit is minute
     MAX_CKPT_THRESHOLD = 900  # seconds
-    MAX_AVG_STEPS = 50
+    MAX_AVG_STEPS = 10  # deprecated
```
BalaBalaYi: still using, delete it if not

dlrover/python/elastic_agent/torch/training.py
```diff
@@ -721,6 +721,12 @@ def _rendezvous(self, worker_group: WorkerGroup) -> None:
             f"{[worker.world_size for worker in workers]}\n"
         )
 
+        if self._diagnose_agent:
```
BalaBalaYi, line 725: remove this condition check?

majieyue: the _diagnose_agent might be null, keep it safer

dlrover/python/common/global_context.py
majieyue: ok

PR #1689, opened 2026-01-27
Enhance logging.

BalaBalaYi:

1. Add new file handler for DLRover framework.
2. Keep and redirect terminal log to specified log directory in torch format when log directory is configured. 
3. Enhance implementations for different torch version compatible.

Enhance logging function.

1. Add a default file handler for DLRover's log using 'DLROVER_LOG_BASE_DIR' env value as log directory. 
2. User can use 'DLROVER_LOG_AGENT_DIR' env variable to specified target log directory. 
When the 'log_dir' is configured, 'redirects' and 'tee' will be set to 'Std.ALL' as default except user give specified value.
The log under DLRover will be:
```
{log_dir}/
  - /attempt_0/
    -  0/
      - stderr.log
      - stdout.log
      - error.json  
```

More details please refer to: [doc](https://github.com/BalaBalaYi/dlrover/blob/37f7f8689d367b75c7ed84a664a3bf95fa1148dc/docs/deployment/logging.md) 

UT and Training with torch 2.1.0 and torch 2.4.0.

### DIFF
--- a/dlrover/python/common/log.py
+++ b/dlrover/python/common/log.py
@@ -87,8 +87,6 @@ def get_logger(
             handlers = [file_handler] + _DEFAULT_HANDLERS
         else:
             handlers = _DEFAULT_HANDLERS
-    elif len(handlers) == 0:
-        handlers = _DEFAULT_HANDLERS
     else:
         handlers.extend(_DEFAULT_HANDLERS)
 


dlrover/python/common/log.py
```diff
@@ -43,14 +44,55 @@ def get_log_level():
...
+            file_handler = RotatingFileHandler(
+                base_log_file,
+                maxBytes=200 * 1024 * 1024,  # 200MB
+                backupCount=3,
+            )
+            file_handler.setFormatter(_DEFAULT_FORMATTER)
+            handlers = [file_handler] + _DEFAULT_HANDLERS
+        else:
+            handlers = _DEFAULT_HANDLERS
+    elif len(handlers) == 0:
```
samplise: Why to process "handlers is None" and "len(handlers) == 0" in different ways?

BalaBalaYi: thx, fixed

PR #1693, opened 2026-02-03
Fix repeated node diagnostitican.

BalaBalaYi:

Use type+index(exclude id) to judge whether there is repeated node.

Bug fix.

No

UT


samplise (approved):
lgtm

PR #1694, opened 2026-02-05
support step type in atorch event logs

majieyue:

Add step type in atorch logs, and dlrover need to apply with it

Only pure train step will detect hang so far

No

UT


dlrover/python/diagnosis/datacollector/atorch_event_collector.py
```diff
@@ -112,6 +112,18 @@ def parse_line(self, line):
...
             event_step = int(text["global_step"])
+            if "step_type" in text:  # backward compatible
+                event_step_type = text["step_type"]
+                if (
+                    event_type == EventTypeName.BEGIN
+                    and event_target == EventTargetName.TRAINER
+                    and event_name == TrainEventName.TRAIN_EVT_STEP
+                    and event_step_type != "train"
+                ):
+                    logger.debug(
```
BalaBalaYi, line 123: use warning?

majieyue: that would be plenty of that, only useful in debug cases

PR #1714, opened 2026-03-19
Support manual resource elastic for allreduce

yifeng-x:

Support dynamic manual adjustment of training task resources (CPU/memory/GPU count) at runtime.

When training exceptions occur due to insufficient resources, the system can automatically adjust resources and restart training, such as in the case of OOM.

No.

Unit test.


workingloong (approved):
Good job.

samplise:
lgtm

PR #1723, opened 2026-06-23
Fix master oom from frequent error message report

samplise:

This pr will fix the oom bug in the master when workers reports large size of error messages in a short-time period.

When a large amount of workers fail at the same time, all of them may report large size of error message to the master and make the master oom.

No.

Unit test.


duck7216 (approved):
lgtm

PR #1730, opened 2026-08-13
Feat: add max_hang_downtime for inspect/eval/first-step hang detection

majieyue:
## Motivation
  The master step-hang detection (`check_job_step_hang`) treats every step
  with a single `hang_downtime` threshold. Legitimately slow steps — the first
  step (compile/warmup) and non-pure-train steps (`train,inspect` / `eval` /
  `train,eval`) — either get falsely flagged or, worse, were silently skipped
  by the collector. The skip itself was buggy: dropping only the `BEGIN` of a
  `train,inspect` step broke the master's strict sequential pairing, which
  deadlocked the whole step-tracking chain at the previous `END`, disabling
  hang detection entirely for the rest of the run.

  ## Changes
  - **New config `--max_hang_downtime`** (default `120` min). Derives
    `max_hang_timeout = max_hang_downtime * 60` (sec), analogous to
    `hang_threshold = hang_downtime * 60`. Floored at `MIN_HANG_DOWNTIME` and
    at `hang_downtime`; also wired into `set_params_from_brain`.
  - **`step_type` is propagated end-to-end** (`AtorchEvent` / `AtorchStepEvent`).
    inspect/eval steps and the first step are no longer dropped — they are
    visible to the master. `check_job_step_hang` now applies `max_hang_timeout`
    to the first step / non-pure-train steps and `hang_threshold` to normal
    pure-train steps. Empty/legacy `step_type` (old agents) is treated as pure
    train (backward compatible).
  - **Forward step-gap tolerance** in `StepEvents.add_step_event`: the BEGIN
    pairing now accepts a forward jump in `step` number and only rejects a
    backward/out-of-order one, so one missing BEGIN (or a skipped/lost step)
    no longer deadlocks the chain — the root cause of the inspect deadlock.
  - **Collector**: removed the `step_type != "train"` BEGIN rejection and the
    `_first_step` skip; `reset_first_step` is now a no-op. `AtorchStepEvent`
    records `is_first_step` (first BEGIN added after a store clear).
  - **`check_ckpt_hang` unchanged**; **`check_event_block` keeps
    `hang_threshold`** — it judges the inter-step gap (a different concern
    from single-step execution; long ckpt/eval/inspect are visible as BEGIN
    and/or suppressed by their own stores).

  ## How to use
  --hang_downtime 5        # normal pure-train step + ckpt threshold (min)
  --max_hang_downtime 120  # ceiling for first step / inspect / eval (min)

  ## Tests
  - `test_training_events.py`: 15/15. Updated `test_atorch_parse_line`
    (inspect BEGIN no longer rejected), `test_atorch_parse_log` (first step is
    now tracked → 3 steps). Added cases for `step_type`/`is_first_step`
    storage (incl. legacy-agent compat), forward-gap tolerance, and
    `max_hang_timeout` threshold selection (first/inspect → 120s, normal train
    → 60s).
  - `test_master_client.py`: updated `test_event_log` (2→3 steps) and
    `test_report_atorch_events` (step_type end-to-end). These need `ray` which
    is absent locally; import failure is pre-existing on `HEAD` and unrelated
    to this change.

### DIFF
--- a/dlrover/dashboard/tests/test_run_dashboard.py
+++ b/dlrover/dashboard/tests/test_run_dashboard.py
@@ -182,6 +182,7 @@ class TestMasterDashboardIntegration(unittest.TestCase):
         mock_args.task_process_timeout = 3600
         mock_args.hang_detection = False
         mock_args.hang_downtime = 300
+        mock_args.max_hang_downtime = 300
         mock_args.pending_fail_strategy = ""
         mock_args.pending_timeout = 0
         mock_args.service_type = ""
@@ -254,6 +255,7 @@ class TestMasterDashboardIntegration(unittest.TestCase):
         mock_args.task_process_timeout = 3600
         mock_args.hang_detection = False
         mock_args.hang_downtime = 300
+        mock_args.max_hang_downtime = 300
         mock_args.pending_fail_strategy = ""
         mock_args.pending_timeout = 0
         mock_args.service_type = ""
@@ -330,6 +332,7 @@ class TestMasterDashboardIntegration(unittest.TestCase):
         mock_args.task_process_timeout = 3600
         mock_args.hang_detection = False
         mock_args.hang_downtime = 300
+        mock_args.max_hang_downtime = 300
         mock_args.pending_fail_strategy = ""
         mock_args.pending_timeout = 0
         mock_args.service_type = ""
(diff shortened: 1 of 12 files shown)


duck7216 (approved):
LGTM

PR #1736, opened 2026-08-21
Feat(master): record nodecheck-failed node host IP before relaunch

majieyue:
## Motivation
  When a node fails the network/node check, the elastic agent reports
  `NodeExitReason.CHECK_FAIL` and the master relaunches it. Today nothing
  prevents the replacement pod (or later pods) from being scheduled back
  onto the same bad physical machine. We want subsequent pods to avoid
  nodecheck-failed nodes via node affinity.

  ## What this changes (upstream side, scheduler-agnostic)
  - `dist_job_manager._relaunch_node`: right before producing the
    replacement pod, if `node.exit_reason == NodeExitReason.CHECK_FAIL` and
    `node.host_ip` is set, record the host IP on the scaler.
    (`node.host_ip` comes from `pod.status.hostIP` via the k8s watcher; on
    hostNetwork pods it equals the physical machine IP.)
  - `Scaler` (base_scaler): owns the accumulated failed-node-IP set plus
    thread-safe `add_failed_node_ip()` / `get_failed_node_ips()`. The base
    scaler is kept free of any scheduler-specific label.

  Only `CHECK_FAIL` relaunches are recorded; OOM / hardware-error / etc. are
  unaffected.

  ## Test
  Adds `test_relaunch_node_records_failed_node_ip` covering:
  - CHECK_FAIL node records its host IP and relaunch still proceeds;
  - non-CHECK_FAIL exit reasons do not record;
  - CHECK_FAIL without a host IP is skipped.

### DIFF
--- a/dlrover/python/master/node/dist_job_manager.py
+++ b/dlrover/python/master/node/dist_job_manager.py
@@ -1175,6 +1175,15 @@ class DistributedJobManager(JobManager):
         return should_relaunch
 
     def _relaunch_node(self, node: Node):
+        # Record the host IP of a nodecheck-failed node before relaunching,
+        # so that the relaunch pod (and later pods) get scheduled away from
+        # it via node affinity in the scaler.
+        if node.exit_reason == NodeExitReason.CHECK_FAIL and node.host_ip:
+            self._scaler.add_failed_node_ip(node.host_ip)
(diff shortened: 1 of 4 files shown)


duck7216 (approved):
LGTM

PR #1737, opened 2026-08-25
Update worker error msg size to 8k

samplise:

Reset the error message size from 1k to 8k.

Previously, we found when many workers failed during a short time period, those workers will report error messages to the master. It may end up with master oom. We limit the error message size to 1k to avoid this problem. However, 1k may miss some important error information. So we set it to be 8k now.

No.

Unit test.

### DIFF
--- a/dlrover/python/elastic_agent/master_client.py
+++ b/dlrover/python/elastic_agent/master_client.py
@@ -42,11 +42,6 @@ from dlrover.python.diagnosis.common.diagnosis_data import DiagnosisData
 from dlrover.python.util.common_util import find_free_port
 from dlrover.python.util.function_util import retry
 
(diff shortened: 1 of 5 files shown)


dlrover/python/elastic_agent/master_client.py
```diff
@@ -45,7 +45,7 @@
 # Truncate error payloads before sending to the master to prevent the master
 # from allocating large gRPC receive buffers and accumulating data in Node
 # objects across relaunches.
-_MAX_ERROR_DATA_LEN = 2048
+_MAX_ERROR_DATA_LEN = 8 * 1024
```
majieyue: why 2 vars? why not define it as global var

majieyue (approved):
LGTM

PR #1721, opened 2026-04-13
Enhance network check for group case

BalaBalaYi:
### What changes were proposed in this pull request?

Implements a multi-phase network check mechanism for node group scenarios (e.g., nodes connected via NVSwitch within a group and via IB across groups), enabling
  more precise fault node localization at both intra-group and inter-group levels.

Core Changes: New GroupNodeNetworkCheckRendezvousManager (rdzv_manager.py)

  A three-phase state machine for network checking:
  - Phase 0 (INTRA_INITIAL): Intra-group adjacent pairing (e.g., G0=[0,1,2,3] → {0,1}, {2,3})
  - Phase 1:
    - If Phase 0 failed → INTRA_DIAGNOSTIC: Intra-group cross pairing (sorted by elapsed time, pairing fastest with slowest to isolate fault nodes)
    - If Phase 0 passed → INTER_INITIAL: Inter-group same-position pairing (e.g., G0=[0,1], G1=[4,5] → {0,4}, {1,5})
  - Phase 2 (INTER_DIAGNOSTIC): Inter-group shifted pairing for further cross-group communication fault diagnosis

### Why are the changes needed?

To enhance network-check on node group situation. 

### Does this PR introduce any user-facing change?

No

### How was this patch tested?

UT and training job on specific case.

PR #1738, opened 2026-08-27
Revert "Enhance network check for group case (#1721)"

majieyue:
Reverts 468d6324.

The hot-upgrade path in servicer._join_rendezvous swaps the live NetworkCheckRendezvousManager for a fresh GroupNodeNetworkCheckRendezvousManager the moment a joined node is observed to have group info, without migrating the already-joined nodes' state (_waiting_nodes / _rdzv_nodes / _rdzv_round / ...).

Because group info reaches the master only via the k8s watch path (pod label scheduling/rack-id), independent of and racing with the join_rendezvous RPC (which carries no group field, comm.py:JoinRendezvousRequest), at scale a large batch of nodes join the base manager BEFORE the watcher syncs their rack-id labels. The first post-sync join then triggers the upgrade and discards all prior joiners. With min_nodes==max_nodes the waiting_timeout branch in _check_rdzv_completed is unreachable, so the new manager never completes -> job hangs in network-check (~join_timeout 5400s) with master logging "Waiting nodes 26/288 lacking ranks 262" forever.

The upgrade approach is structurally unsound: it decides a job-level policy (group vs non-group) by observation on a stateful hot path, swaps a stateful object mid-flight, and lets agent and master derive "is group job" from two different signals (env vs pod-label-watch). Reverting to redesign the group network check by selecting the manager type at construction with a single source of truth for check_round.

Please describe the changes you have made or proposed in this pull request.

Specify whether this pull request introduces any changes that users will directly interact with or notice.

Detail the testing process you have undertaken to ensure the changes in this pull request are valid and working as intended.


duck7216 (approved):
LGTM

PR #1740, opened 2026-08-31
Feat(master): ep_pp_dp node-group scheduling for MoE segment placement

majieyue:
## What
  Adds --node-group-strategy (contiguous | ep_pp_dp) plus --tp/--pp/--ep/--cp master args, so dlrover places Megatron MoE worker nodes onto node groups (== physical segments) with EP groups and PP kept
  intra-segment while only the dense DP allreduce crosses segments.

  ## Why
  Default ep-dp-pp + contiguous scheduling makes PP (latency-critical, sequential) span all segments — worst case. Megatron forces Order.endswith(pp) when EP>1 (parallel_state.py assert) and NCCL
  --use-tp-pp-dp-mapping is incompatible with EP, so the fix lives at the scheduling layer.

  ## How
  group(node_rank) = (node_rank % DP_nodes) // (DP_nodes // G), DP_nodes = N/(TP*PP*CP). Each pipeline stage is striped across all G segments identically, so adjacent PP stages same-position nodes share a
  segment. validate_topology fails fast on TP/CP>1, dense_dp%G!=0, node not segment-aligned, EP not dividing a segment-per-stage block, EP not whole-nodes, unequal group sizes, group ids != {0..G-1}.
  resolve_group_id takes an optional NodeGroupSchedule; None/contiguous keeps existing behavior.

  ## Backward compat
  Unset / contiguous (default) unchanged; parallel sizes read and validated only under ep_pp_dp.

  ## Tests
  New test_node_group_strategy.py (24): stripe formula, every validate_topology constraint pass/fail, rank-level EP/PP-intra & DP-cross invariants on the 9216-card job (N=1152,R=8,TP=1,PP=16,EP=32,G=9)
  plus smaller topologies. Extended test_args, test_args_util, test_pod_scaler, test_job_manager.

PR #1741, opened 2026-09-02
Feat(master): validate reported local_world_size vs ep_pp_dp ranks_per_node

majieyue:
## Summary

  Follow-up to #1740, implementing the review-agreed remediation for the
  `ranks_per_node` inference issue: `ranks_per_node` is inferred from the
  requested GPU count at master startup, while the actual per-node process
  number comes from `dlrover-run --nproc_per_node` on the trainer side and
  only reaches the master when an agent joins the rendezvous (stored by
  `rdzv_manager` as `process_num`). This PR re-validates the two against
  each other at join time, so a misconfigured job fails fast instead of
  training with wrong rank boundaries.

  ## Changes

  - `RendezvousManager` carries an expected local world size, injected by
    `DistributedJobMaster` **only when the ep_pp_dp strategy is enabled**
    (`None` disables the check, so NetworkCheck rendezvous and non-ep_pp_dp
    jobs are untouched).
  - `join_rendezvous` re-validates each node's reported `local_world_size`
    **before mutating any state**: on mismatch it logs an ERROR with a fix
    hint and raises `ValueError`, so the job fails before the world freezes
    (e.g. 8 requested GPUs with `nproc_per_node=4`: startup validation
    passes with S=16 while the real S=8 makes EP groups cross segments).
  - `DistributedJobManager.get_expected_ranks_per_node()` exposes the
    validated R for the injection.

  ## Tests

  - `test_rdzv_manager`: a node with the expected nproc joins normally; a
    mismatching node raises with a clear message, leaves no waiting state,
    and the ERROR log carries the `--nproc_per_node` hint.
  - `test_node_group_strategy`: worker-level invariants on the 9216-card
    job (N=1152 = 9 segments × 128 nodes, R=8, PP=16, EP=32) — every
    segment holds exactly 128 workers on the same per-stage slot; every EP
    world (4 workers) and every PP world (the same 4-worker slot across
    all 16 stages, e.g. `ep_pp0 = 0-3, 72-75, ..., 1080-1083`) stays
    inside a single segment.
  - `test_job_manager`: `get_expected_ranks_per_node` returns the
    validated R only for ep_pp_dp.

  ## Notes

  - This branch is **stacked on #1740** (it uses `NodeGroupSchedule` from
    there), so the diff currently includes #1740's commits as well and
    will collapse to this commit's changes once #1740 merges.
  - Remaining agreed remediation will follow in separate PRs: an explicit
    `--ranks-per-node` master argument, and fixing the `gpu_num` parsing
    for limits-only / non-NVIDIA resources.

### DIFF
--- a/dlrover/python/master/dist_master.py
+++ b/dlrover/python/master/dist_master.py
@@ -152,6 +152,15 @@ class DistributedJobMaster(JobMaster):
             RendezvousName.TRAINING: create_training_rdzv_manager(),
             RendezvousName.NETWORK_CHECK: NetworkCheckRendezvousManager(),
         }
+        if self.job_manager is not None:
+            # The ep_pp_dp ranks_per_node is inferred from the requested
+            # GPU count at startup; re-validate it against the actual
+            # local_world_size reported by trainers at rendezvous time.
+            expected_ranks = self.job_manager.get_expected_ranks_per_node()
+            if expected_ranks is not None:
+                self.rdzv_managers[
+                    RendezvousName.TRAINING
+                ].set_expected_local_world_size(expected_ranks)
         self.diagnosis_manager = DiagnosisMaster(args)
         self._event_reporter = get_event_reporter()
         self.job_metric_collector = self._create_metric_collector_if_needed(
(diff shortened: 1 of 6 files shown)


samplise (approved):
lgtm

PR #1747, opened 2026-09-11
Feat(scheduler): add k8sClient.patch_images_to_pod for in-place image update

majieyue:
## What

Add `k8sClient.patch_images_to_pod(name, images)` which patches the images
of named containers in a Pod via `patch_namespaced_pod`. The kubelet
updates the container in place, so the Pod keeps its name, IP, labels and
env.

## Why

The internal Ant master (dlrover-addon) watches the training job CRD.
When a user updates the worker image, the master first refreshes the
in-memory JobResource image (so later relaunch/scale-up also uses the new
image) and then patches every worker Pod's main container in place
instead of recreating the Pods. This upstream method is the k8s API
primitive that flow needs; it mirrors the existing
`patch_labels_to_pod` / `patch_annotations_to_pod` pattern
(`@retry_k8s_request` + `patch_namespaced_pod`).

## Implementation

- `patch_images_to_pod(name, images: Dict[str, str])`: the body only
  touches `spec.containers` images listed in the `images` mapping
  (container name -> image), so init containers and sidecars are left
  untouched.
- Unit tests in `test_k8s.py`: the success path asserts the exact patch
  body and namespace; the failure path asserts the retry-exhausted —>
  `None` behavior of `@retry_k8s_request`.

### DIFF
--- a/dlrover/python/scheduler/kubernetes.py
+++ b/dlrover/python/scheduler/kubernetes.py
@@ -294,6 +294,30 @@ class k8sClient(Singleton):
         )
 
     @retry_k8s_request
+    def patch_images_to_pod(self, name, images: Dict[str, str]):
+        """Patch images of named containers in a Pod.
+
+        The Pod is not recreated: the kubelet updates the container
+        image in place, so the Pod keeps its IP, labels and env.
+
+        Args:
+            name: str, the pod name.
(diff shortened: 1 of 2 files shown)


duck7216 (approved):
LGTM

PR #1748, opened 2026-09-17
Feat(master): require node-group-strategy and add expert tensor paral…

majieyue:
---
  Summary

  Three parameter-semantics changes for node-group scheduling (both apply at job-manager init, so a violation fails the master start cleanly):

  1. --node-group-strategy defaults to None (was "contiguous"), and the contiguous strategy is renamed ep_dp_pp with unchanged placement semantics. Configuring --group-affinity or --soft-group-affinity
  now requires the strategy to be set explicitly; conversely, a strategy without any group affinity is rejected.
  2. New --expert-tensor-parallel-size (alias --etp): when unset it inherits --tensor-model-parallel-size; when set explicitly it must be used together with --expert-model-parallel-size. The real EP group
  size becomes ETP * EP wherever EP previously sized a group: topology constraints C/E, the soft-group EP alignment (ep_workers = EPG/R), the dense-dp divisibility, the ep_pp_dp column-packing slot, and
  the group-range logging.
  3. --tp / --cp are recorded but no longer used: the TP==1/CP==1 rejections are removed and the layout math only involves PP — numerically identical for every topology that passed the old validation.

  Motivation

  Megatron MoE jobs run experts with an expert-tensor-parallel dimension, so the expert communication group actually spans ETP*EP ranks, not EP. Sizing the EP slot by EP alone mis-aligned the segment
  affinity for such topologies. The inherit-from-TP default keeps single-dimensional MoE jobs unchanged.

  Compatibility notes

  - Jobs with no node-group parameters: zero behavior change.
  - --node-group-strategy=contiguous must be renamed to ep_dp_pp; contiguous is now rejected by argparse, and omitting the strategy while configuring an affinity is rejected at master start (loud failure
  replaces the old silent default).
  - Footgun worth knowing: an unset --etp inheriting --tp means --tp=2 alone scales the EP group to 2*EP once an affinity is configured — pass --etp=1 to keep --tp purely recorded.

  Tests

  9 new cases covering the strategy-required/pairing errors, ETP inheritance, EPG-scaled constraints, and recorded-only TP/CP; all related suites pass (the 5 test_job_manager.py failures are
  order-dependent and reproduce on unmodified HEAD).

  Companion change: the Ant-internal dlrover-addon fork mirrors the new flag definitions and pass-through (AntCode PR 927 on Arc/dlrover-addon).


samplise (approved):

duck7216 (commented):
LGTM

PR #1749, opened 2026-09-20
Fix(master): clear ckpt/eval step events on restart boundaries to sto…

majieyue:
Description:

  Problem

  We hit a production incident where a worker was restarted midway through a checkpoint save (pod evicted on a hardware fault). Hang detection correctly identified the dangling save (step hang is False,
  ckpt hang is True) and restarted the worker via RESTART_WORKER. After that, the job was killed every ~30 minutes forever — each recovery window of low tensor utilization re-triggered the same
  misdetection, and the job never finished self-healing.

  Root cause

  Two defects compound on the restart boundary:

  1. All restart boundaries clear only train_steps. TrainingHangDiagnostician.resolve() (DO_FAILOVER), DiagnosisMaster.pause_observing() and continue_observing() (the rendezvous path every worker restart
  goes through) all call train_steps.clear_step_events() only. The dangling ckpt BEGIN reported by the dead session survives every restart in JobEventContext.ckpt_steps with an ever-growing now -
  localtime, so check_ckpt_hang() stays True permanently. RESTART_WORKER restarts training processes in place — the master process and its event state are never reset.
  2. A dangling BEGIN blocks all subsequent ckpt/eval events. _add_generic_event() rejects every BEGIN while the latest event is not in END state (invalid ckpt step), and rejects every END whose step
  doesn't match the dangling BEGIN. The state can never be superseded by the new run's own save events, so even a healthy recovery cannot clear the misdetection.

  The step-hang path already tolerates recovery after a restart (first step after a clear is checked against max_hang_timeout), but the ckpt-hang path carried pre-restart state with zero grace. The kill
  cadence is exactly hang_downtime: pause/continue_observing reset the metric history, so the tensor-util drop window re-accumulates hang_downtime minutes during recovery, at which point ckpt_hang
  (instantly True from the stale BEGIN) fires the restart.

  Changes

  - Add JobEventContext.clear_all_step_events() clearing train/predict/ckpt/eval queues together, and call it at all three restart boundaries (training_hang.py resolve, diagnosis_master.py
  pause/continue_observing) so a restart starts a fresh observation session (consistent with these sites already clearing node metrics).
  - _add_generic_event(): supersede a dangling BEGIN when a newer BEGIN of a different step arrives (logged + replaced). This also heals the case where a save END is simply lost (parse failure / transient
  report loss) with no restart involved.

  Compatibility notes

  - Same-step BEGINs arriving while a BEGIN is pending are still rejected, preserving the existing duplicate-re-report semantics covered by test_atorch_ckpt_event / test_atorch_eval_event (the arriving
  END closes the original event).
  - train_steps handling (including the missing-BEGIN forward-gap tolerance) is unchanged; predict_steps is also cleared but unused by hang detection.

  Testing

  - New test_ckpt_hang_restart_boundary: dangling ckpt BEGIN judged as hang → boundary clear → no hang; fresh save of the new run registers and closes normally.
  - New test_ckpt_event_supersede_dangling_begin: not-newer BEGIN rejected; newer different-step BEGIN supersedes; its END closes; no hang.
  - Full test_training_events.py (17 tests) and test_diagnosis_master.py + test_diagnostician.py (20 tests) pass; pre-commit (ruff check/format, mypy, copyright) clean.


duck7216 (approved):
LGTM

samplise (approved):

PR #1750, opened 2026-09-21
Feat(master): rerank the rendezvous world by the psw network topology

majieyue:
## Background

Training nodes spread across multiple physical switch groups (psw). Ranks are
currently assigned by pod creation order, regardless of the network topology,
so EP/TP/PP communication straddles switches heavily and the collective
bandwidth is capped by the inter-switch links.

This PR adds `--enable-topology-rerank`: at the elastic-training rendezvous
the master observes every node's **psw** from a cluster topology JSON, then
**reorders the frozen rendezvous world** so that EP/PP communication groups
stay intra-psw as much as possible. The network-check rendezvous is NOT
reordered.

## Core design: reorder the world order, never renumber node_rank

Worker processes derive all ranks (RANK / GROUP_RANK / ...) from the world
dict order returned by `get_comm_world` each rendezvous round — `node_rank`
is only the stable pod identity. Therefore **changing the world dict order
is enough to rerank every process**, with zero agent / gRPC protocol changes.
The world travels through the pickle-based `comm.Message` channel (not a
proto map), so dict insertion order survives the RPC.

## What's inside

All rerank logic is in the new module
`dlrover/python/master/elastic_training/topology_rerank.py`:

- **`GroupTopologyQuerier`** — loads the cluster topology
  `{node_ip: psw}` from the env `DLROVER_GROUP_TOPOLOGY` (inline JSON) or
  the file at `DLROVER_GROUP_TOPOLOGY_FILE` (inline env wins).
- **`GroupTopologySorter`** — one node group per distinct psw, arbitrary
  group sizes (no EP-slot alignment constraint), laid out by the configured
  `--node-group-strategy`:
  - `ep_dp_pp`: groups occupy contiguous world ranges in ascending psw order;
  - `ep_pp_dp`: EP slots are greedily packed into ep-pp columns so whole
    PP columns stay intra-psw (only residual fragments share a column);
    leftover whole slots and sub-slot crumbs are appended and reported.
  - Group members are instantiated in ascending node_rank order →
    same metas always produce the same world order (idempotent across
    rendezvous rounds).

The querier/sorter are **injected into the training rendezvous manager only
when the flag is on**; the default behavior and the whole
`--soft-group-affinity` / `--group-affinity` code paths are unchanged
(`--enable-topology-rerank` is mutually exclusive with both, validated at
master startup; it requires `--node-group-strategy`).

### Edge condition: unknown node IP

A node whose IP is not in the topology JSON reports an empty psw and joins
the **DEFAULT group**, a peer of the known groups (group ids follow ascending
psw, `""` sorts first; no tail-demotion). In `ep_dp_pp` it takes the first
contiguous range; in `ep_pp_dp` it packs into the greedy columns like any
other group. It shows as `<default>` in the logs. Everything else is
unchanged.

### Observability

Every rerank step is logged with a greppable `RERANK` prefix on both the
master (config / per-psw-group / per-node `old->new` world position / final
order / EP-slot straddle stats) and the agent (this node's position and the
per-worker rank before/after each round).

## Companion fix

Flash-checkpoint backup-rank grouping (`CkptReplicaManger._get_backup_ranks`)
now keys nodes by the **world position** (`self.rank // local_world_size`)
instead of the `node_rank` value, so backup groups stay mutually consistent
after the reorder. No behavior change when the flag is off.

## Usage

```bash
dlrover-run ... # master side:
--enable-topology-rerank --node-group-strategy=ep_pp_dp \
--expert-model-parallel-size=8 --expert-tensor-parallel-size=1 ...
# platform feeds the topology (either way):
export DLROVER_GROUP_TOPOLOGY='{"33.167.120.39": "SG2", ...}'
# or: DLROVER_GROUP_TOPOLOGY_FILE=/path/topology.json
```

## Tests

- New `test_topology_rerank.py` (querier loading incl. bad JSON / both
  layouts incl. default group, straddles, determinism / idempotence) +
  extended `test_rdzv_manager`, `test_args`, `test_job_manager`,
  `checkpoint_backup_test`.
- python/trainer test suites match the master baseline (failures are
  pre-existing local-env ones, verified item-by-item against a master
  checkout).

## Notes

- `--enable-topology-rerank` ⊥ `--group-affinity` / `--soft-group-affinity`
  (hard error when combined); requires `--node-group-strategy`.


duck7216 (approved):
LGTM

